Issue #6, opened 2018-06-25
Website Discord integration


Astraeus-:
Automatically update roles on Discord for captain status changes and registration as Free Agent.

Task list

- [x] Store our users Discord Id snowflakes in our database
- [x] Integrate storing users Discord Id snowflakes into registration flow
- [x] Update current user-base to include Discord Id snowflake
- [x] [Complete Discord role assignment code](https://github.com/Fabian-Sommer/HeroesLounge/blob/master/plugins/rikki/heroeslounge/classes/discord/RoleManagement.php)
- [x] Integrate Discord role assignment code with current flows

PR #16, opened 2018-08-07
Discord integration

Astraeus-:
Started work on Discord integration with the website #6 

I don't think it's complete and I haven't been able to test it very well. There are still a few things that need to be considered and altered before it's ready for live deployment. I believe I caught all of the breaking changes that I made in discord/Attendance correctly though.

- What to do when the Discord role assignment/removal is the fail point in a website flow. 
e.g: When creating a team, the API is not available and/or the account isn't in our Discord server etc.
- There will likely be old users that are no longer on our Discord server, or of whom the Discord Tag is no longer correct. These users won't be migrated correctly and I haven't added any testing for their discord_tag not existing.
- Not sure if my migrations are 100% correct. I also don't use the code I wrote to populate the discord_id field for our existing users.

### DIFF
--- a/plugins/rikki/heroeslounge/classes/discord/Attendance.php
+++ b/plugins/rikki/heroeslounge/classes/discord/Attendance.php
@@ -1,6 +1,5 @@
 <?php namespace Rikki\Heroeslounge\classes\Discord;
 
-  
 use Rikki\Heroeslounge\Models\Sloth as SlothModel;
 use Rikki\Heroeslounge\classes\Discord\AuthCode;
 use Db;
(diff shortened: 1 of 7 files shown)


plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -9,55 +8,55 @@ class Attendance
...
+              }
+          }
+      }
+
+      return $presentUsers;
     }
 
     public function CheckAttendance()
     {
         $discordTags = Db::table("rikki_heroeslounge_sloths")->where("is_captain", 1)->whereIn('team_id', Db::table('rikki_heroeslounge_team_division')->whereIn('div_id', [1,2,3,4,5])->lists('team_id'))->lists("discord_tag");
```
Fabian-Sommer: This is both bad style (do not use Db::table, instead go for Sloth::where("is_captain", 1)->... and also wrong, as it would only get you captains of teams that were in Div 1 through 5 in Season 4 - do not hardcode the division ids.
I am also unsure what this function is supposed to achieve - get attendance of all captains of all teams in the current season? Rather check for all teams of all divisions of all currently active seasons.

plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -68,8 +67,82 @@ public function CheckAttendance()
...
+        if ($presentUsers) {
+          foreach ($presentUsers as $user) {
+            $userTag = $user["username"] . "#" . $user["discriminator"];
+            if ($userTag == $discordTag) {
+              $userId = $user["id"];
+              break;
+            }
+          }
+        }
+        return $userId
```
Fabian-Sommer: The way this function is named it should return either true or false. Rename to GetUserId or GetDiscordUserId otherwise.

plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -68,8 +67,82 @@ public function CheckAttendance()
...
+      $url = 'https://discordapp.com/api/guilds/200267155479068672/members/';
+
+      $auth_header = AuthCode::getCode();
+      $headers = [
+        "Content-Type:application/x-www-form-urlencoded",
+        $auth_header,
+        "User-Agent: HeroesLounge (http://heroeslounge.gg, 0.1)"
+      ];
+
+      $ch = curl_init($url . $discordId);
```
Fabian-Sommer: Better to concatenate $discordId to url directly

plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -68,8 +67,82 @@ public function CheckAttendance()
...
+
+      return $presentDiscordTags;
+    }
+
+    /*
+      Unsure if this is being done correctly and if it should not instead be inside of the migration file?
+    */
+    public function migrationDiscordTagsToIds ()
+    {
+      $sloths = Db::table('rikki_heroeslounge_sloths')->select('id', 'discord_tag');
```
Fabian-Sommer: Again, style. Do Sloth::all() and access individual fields with $sloth->id and $sloth->discord_tag

plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -68,8 +67,82 @@ public function CheckAttendance()
...
+      $presentUsers = $this->FetchUsers();
+
+      if ($presentUsers) {
+        foreach ($presentUsers as $user) {
+          $userTag = $user["username"] . "#" . $user["discriminator"];
+          foreach ($sloths as $sloth) {
+            if ($userTag == $sloth["discord_tag"]) {
+              Db::table('rikki_heroeslounge_sloths')
+              ->where('id', $sloth["id"])
+              ->update(['discord_id' => $user["id"]]);
```
Fabian-Sommer: $sloth->discord_id = $user["id"]; $sloth->save();

plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -68,8 +67,82 @@ public function CheckAttendance()
...
+
+      foreach ($users as $user) {
+        $presentDiscordTags[] = $user["username"] . "#" . $user["discriminator"];
+      }
+
+      return $presentDiscordTags;
+    }
+
+    /*
+      Unsure if this is being done correctly and if it should not instead be inside of the migration file?
```
Fabian-Sommer: Ok here

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -182,11 +182,14 @@ public function onRegister()
...
+            /*
+             * Check if the user is on the Heroes Lounge Discord server.
+             */
 
             $DiscordAttendance = new Discord\Attendance;
-            $matchFound = $DiscordAttendance->IsOnServer($data['discord_tag']);
+            $userDiscordId = $DiscordAttendance->IsOnServer($data['discord_tag']);
 
-            if (!$matchFound) {
+            if (empty($userId)) {
```
Fabian-Sommer: See above, should be true/false

plugins/rikki/heroeslounge/controllers/Team.php
```diff
@@ -47,11 +49,16 @@ public function formAfterUpdate($model)
             $team->sloths->each(function($sloth){
                 if($sloth->is_captain)
                     $sloth->is_captain = false;
+                    /*
+                      Requires testing of existance of discord_id in the database.
+                    */
+                    $DiscordRoleManagement = new Discord\RoleManagement;
+                    $DiscordRoleManagement->UpdateUserRole("PUT", $sloth->discord_id, "Captains");
```
Fabian-Sommer: I think this should be DELETE instead of PUT

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -255,15 +259,24 @@ public function onRegister()
...
     {
+      $DiscordRoleManagement = new Discord\RoleManagement;
         try {
             foreach ($this->seasons as $season) {
                 if ($season->current_round == 0) {
                     $val = post('part_seas_'.$season->id);
                     if (isset($val) && !$season->free_agents->contains($this->sloth->id)) {
                         $season->free_agents()->attach($this->sloth->id);
+                        /*
+                        Requires testing of existance of discord_id in the database.
```
Fabian-Sommer: You may assume existence

Fabian-Sommer: -> (please delete these comments)

plugins/rikki/heroeslounge/classes/discord/Attendance.php
Astraeus-: I'm unsure a lot of the code in the Attendance.php file was written by Epixors at the time. I'm unsure if he also wrote this function as well

plugins/rikki/heroeslounge/classes/discord/Attendance.php
Astraeus-: Made some changes, unsure if 100% correct now.

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -187,9 +187,9 @@ public function onRegister()
              */
 
             $DiscordAttendance = new Discord\Attendance;
-            $userDiscordId = $DiscordAttendance->IsOnServer($data['discord_tag']);
+            $userDiscordId = $DiscordAttendance->GetDiscordUserId($data['discord_tag']);
 
-            if (empty($userId)) {
+            if (empty($userDiscordId) == true) {
```
Fabian-Sommer: == true is generally discouraged, since unnecessary and less readability

plugins/rikki/heroeslounge/classes/discord/Attendance.php
Fabian-Sommer: I do not even know if this function is being used anymore, or by what. Might just be some legacy code lying around.

Issue #30, opened 2018-10-19
Show NA/EU in calendar


Fabian-Sommer:
The calendar ("Upcoming matches" component) should show whether a match belongs to NA or EU

PR #32, opened 2018-10-20
Add region tag to "upcoming matches" calendar component

stephwang:
This change adds a tag in front of the division title to indicate EU vs. NA by using the teams' region_ids.
Addresses issue #30


stephwang:
Screenshot of how this looks 
region cal

PR #33, opened 2018-10-21
Updated region assignment for upcoming matches

stephwang:
Check the tournament region first, which may assign regions to matches with TBD teams. If no region assigned in the tournament/season, use the teams' regions to determine the proper region tag.

Dev screenshot - regions are being pulled in for all matches now
region cal


stephwang:
Related to issue #30

Issue #38, opened 2018-11-14
Limit size of images


Fabian-Sommer:
We use TinyPNG for compression of images, which fails to do so once they are >100MB. Size of all images that can be uploaded (profile and team pics) needs to be limited accordingly.

Astraeus-:
Currently some users have used .svg files for their team logo. We're going to need to be more strict about only supporting the .png file format as tinyPNG only supports png images. (Also OBS struggles with .svg files for our casters too)

PR #47, opened 2018-11-24
Add client-side validation to image uploads

stephwang:
Related to #38 

This diff adds an image validator function within `selectFile.js`. The JS file is shared by all pages that contain a file uploader, and adds an event handler on change to image uploaders. The validator will raise errors and clear the uploader field if a user attempts to upload a non-PNG image or attempts to upload a file >100MB in size.

We also add the extra attribute `accept="image/png"` to all logo, avatar, and banner uploaders to limit the file selection to PNG by default. We also add some helpful text under each of these fields to let users know the upload restrictions.

This is what the error looks like after a user attempts to upload an invalid file:
capture

### DIFF
--- a/plugins/rikki/heroeslounge/assets/js/selectFile.js
+++ b/plugins/rikki/heroeslounge/assets/js/selectFile.js
@@ -23,6 +23,34 @@ jQuery(function() {
           }
 
       });
+
+      // Client-size checks for image uploaders
+      // Images must be PNG and under 100MB 
+      jQuery("input[type='file'][accept='image/png']").on('change', validateImageFileAndSize);
+
+      function validateImageFileAndSize() {
+          if (this.files.length === 0) {
+              return;
+          }
+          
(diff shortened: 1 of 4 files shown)


Fabian-Sommer:
I dont see the restriction of <=1280x1280 enforced anywhere (or are PNG of such size always >100MB?), but those images are never displayed in such a huge size anyway, so people might as well upload small images.
Looks good otherwise

stephwang:
For logo uploads we have [server-side checks](https://github.com/Fabian-Sommer/HeroesLounge/blob/master/plugins/rikki/heroeslounge/components/ManageTeam.php#L103) on the dimensions, so I figured I'd surface that in the upload requirements.
I can add some client-side checks for it as well if we want. Dimensions checks seemed trickier to do client-side and we wouldn't be able to make use of one shared validation function, so I skipped it for now.

PR #55, opened 2019-01-10
Fixes updating roster for teams not participating in a season

### LINKED ISSUE #53: Teams updating roster during seasons
Currently it's not possible for any team to update their rosters on the website when there is a season running.

It would be a lot nicer if only rosters of teams participating in an active season or tournament are locked. This would allow people to form and prepare their team if they join heroeslounge during a running season.

Astraeus-:
Minor fix on oversight when the roster code was reformatted on creation of the captain promotion. Fixes #53


Fabian-Sommer:
Issue has been resolved

PR #71, opened 2019-03-05
Add some install step echoing

rogusdev:
Helps reduce confusion and accidents

### DIFF
--- a/install_hl.sh
+++ b/install_hl.sh
@@ -3,19 +3,27 @@
 BASE=/opt/lampp
 HTDOCS=$BASE/htdocs
 
+if [ ! -f /vagrant/hl.sql.zip ]
+then
+  echo "db zip must be present in /vagrant to run this!"
+  exit
+fi
+
+echo 'Deleting, moving, symlinking'
 sudo rm -f $HTDOCS/hl/install.php
 sudo rm -rf $HTDOCS/hl/install_files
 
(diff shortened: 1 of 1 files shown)


install_hl.sh
```diff
@@ -3,19 +3,27 @@
 BASE=/opt/lampp
 HTDOCS=$BASE/htdocs
 
+if [ ! -f /vagrant/hl.sql.zip ]
+then
+  echo "db zip must be present in /vargrant to run this!"
```
Astraeus-: I take it this should be `/vagrant` instead of `/vargrant`?

install_hl.sh
```diff
@@ -3,19 +3,27 @@
...
 sudo -u daemon ln -s /vagrant $HTDOCS/hl
 
+echo 'Copying (takes a long while)'
 sudo cp /vagrant/README.md $HTDOCS/install-master/
 sudo cp /vagrant/.gitignore $HTDOCS/install-master/
 
 sudo -u vagrant cp -r $HTDOCS/install-master/. $HTDOCS/hl/
 sudo rm -rf $HTDOCS/install-master/
 
+echo 'Importing db (a short while)'
```
Fabian-Sommer: Maybe this should be a complete sentence to avoid confusion
```suggestion
echo 'Importing db (takes a small amount of time)'
```

rogusdev:
@Astraeus- updated

PR #80, opened 2019-03-15
Fix back-end search for season and match creation

Astraeus-:
I believe this fixes searching for teams in the back-end when adding teams to a season or a new match.

Also in local testing, setting a recordsPerPage did improve load time when opening the team list.

Not entirely sure about my choice of name for the team list, as it's basically a smaller version of the original `columns.yaml` file on the team model.

### DIFF
--- a/plugins/rikki/heroeslounge/controllers/match/config_relation.yaml
+++ b/plugins/rikki/heroeslounge/controllers/match/config_relation.yaml
@@ -10,6 +10,9 @@ teams:
         toolbarPartial: teams_toolbar
     manage:
         showSearch: true
+        recordsPerPage: 20
+        list: $/rikki/heroeslounge/models/team/relation_columns.yaml
+            
 games:
     label: Games
     view: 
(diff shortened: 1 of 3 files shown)


plugins/rikki/heroeslounge/controllers/season/config_relation.yaml
```diff
@@ -39,3 +39,5 @@ teams:
         toolbarPartial: teams_toolbar
     manage:
         showSearch: true
+        recordsPerPage: 20
+        list: $/rikki/heroeslounge/models/team/small_list.yaml
```
Fabian-Sommer: Please end files with a newline, it slightly messes up git history once anybody else adds more lines otherwise

plugins/rikki/heroeslounge/controllers/match/config_relation.yaml
```diff
@@ -10,6 +10,9 @@ teams:
         toolbarPartial: teams_toolbar
     manage:
         showSearch: true
+        recordsPerPage: 20
+        list: $/rikki/heroeslounge/models/team/small_list.yaml
```
Fabian-Sommer: Please rename small_list to something like relationship_columns.yaml

PR #84, opened 2019-03-22
Now fetches the new discordtag from Discord

Astraeus-:
Instead of having the user manually enter their new discord tag, they're now able to press a "sync" button, which will retrieve their new Discord tag automatically. This helps to ensure that Discord tags are guaranteed to be more correct and fixes users entering false values.


Astraeus-:
Also limited the migration to users that don't have a discord id defined on their user. We could automatically check every week if users have changed their discord tag, but that seems like a lot of unneeded server load.

rogusdev:
Is there a way to setup localdev to contact discord for testing these sorts of things?  I assume I need to put my discord client id + secrets somewhere?  Would love to get https://github.com/Fabian-Sommer/HeroesLounge/wiki/Discord-integration filled in

plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -126,6 +126,36 @@ public static function migrateDiscordTagsToIds()
           }
         }
       }
+    }
+
+    public static function getDiscordTag($discordId)
```
rogusdev: how do we get the discord tag already?  is this functionality not already present in some form in the codebase?

Astraeus-: We currently get a user's Discord tag by inefficiently checking what they fill in on registration against every single member in our Discord server. Because we now also store a discord id, which is unique to an account, we're able to directly query Discord for user information.

plugins/rikki/heroeslounge/components/slothaccount/update.htm
```diff
@@ -238,7 +245,7 @@ <h3>Description</h3>
                     </div>
                     <input class="form-control" placeholder="BattleTag" title="Battle Tag - talk to a moderator on Discord if this has changed" type="text" name="battle_tag" disabled value="{{ __SELF__.sloth.battle_tag }}">
                 </div>
-                <div class="input-group col-lg-2 mb-2 col-md-3">
+                <div class="input-group col-lg-2 mb-3 col-md-4">
```
Fabian-Sommer: Changing mb-2 to mb-3 here should lead to an additional row with very little content. Was there too little space available?

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -461,6 +460,39 @@ public function onUpdateLinks()
...
+                    return $redirect;
+                }
+            } else {
+                Flash::error(`Could not update Discord tag!`);
+                if ($redirect = $this->makeRedirection()) {
+                    return $redirect;
+                }
+            }
+        } else {
+            Flash::error(`Could not update Discord tag, please contact a moderator for assistance.`);
```
Fabian-Sommer: Please allow users who do not have a valid id to just enter their Discord tag on their own. We do not want additional workload for mods when aviodable.

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -461,6 +460,39 @@ public function onUpdateLinks()
...
+            if (!empty($newTag)) {
+                $sloth->discord_tag = $newTag;
+                $sloth->save();
+    
+                Flash::success('Discord tag was updated successfully!');
+                if ($redirect = $this->makeRedirection()) {
+                    return $redirect;
+                }
+            } else {
+                Flash::error(`Could not update Discord tag!`);
```
Fabian-Sommer: This is not a very helpful message for the user. You can add something like 'Please try again later'

plugins/rikki/heroeslounge/components/slothaccount/update.htm
Astraeus-: This straightened out the bottom row of updating the user social information, as (almost) every other social "box" had this definition for it's size.

plugins/rikki/heroeslounge/components/SlothAccount.php
Astraeus-: I feel like this will cause confusion for users that will instead try to manually enter their Discordtag instead of having us do the update. I'm quite certain that you're also aware how much our users already struggle getting their new tag right on registration, let alone if they try so here, without verification.

plugins/rikki/heroeslounge/components/slothaccount/update.htm
Fabian-Sommer: Having them being the same size is fine. You should also aim to fill the rows and not have a single dangling small element that gets it own row.

Both goals can be accomplished at the same time here by changing the sizes of other elements

Astraeus-: I agree, so I'm a bit confused about your statement. I tested both in Firefox and Chrome and do not experience a single dangling element on it's own row. This change only intends to align the boxes on the second row better, which it does in my testing.

plugins/rikki/heroeslounge/components/SlothAccount.php
Fabian-Sommer: I trust users are able to enter their own discord tag. Feel free to verify it and display a message if they are not found on the server. If all fails, mods can still help out.

Issue #15, opened 2018-08-04
Display long title for divisions in backend relations instead of title


Astraeus-:
**Division selector drop-down**
![backend create match division selector](https://user-images.githubusercontent.com/6303961/43677020-a02b4b5c-97fb-11e8-92eb-7a551c9853eb.PNG)

Associating a match with a division or playoff group, requires you to select the correct one out of multiple similarly named divisions or groups. Specifying a season would resolve this problem for regular season games.

However for Championship and Cup matches this wouldn't solve much. E.g: You'd still need to guess which "Group A" belongs to the Championship or the Cup.
Potential future tournaments will also not have a season that they're associated with and probably run into similar problems as mentioned above with the groups from the Cup and the Championship.

Fabian-Sommer:
May or may not be possible with the framework

rogusdev:
For starters on creating a match, could you change `nameFrom: title` in plugins\rikki\heroeslounge\models\match\fields_on_create.yaml to something like the new `nameFrom: overview_display_title` and it would do what you want?  As in, will that be populated appropriately?  Or is there some way to reference a calculated field on the model rather than a db field?  And the calculated field be some method on the model plugins\rikki\heroeslounge\models\Match.php somehow?

Astraeus-:
Changing to `nameFrom: overview_display_title` would not work as not every division has this defined. You'd get a long list of blanks with two named `Standings` for the division S divisions. The [OctoberCMS Documentation](https://octobercms.com/docs/backend/forms) has the various things that would be possible. I think the best way to go about this is to probably create a custom dropdown instead of relying on the relation.

Issue #86, opened 2019-03-23
On the edit match page, the participating teams are not listed anywhere


rogusdev:
e.g.
http://localhost:8080/hl/backend/rikki/heroeslounge/match/update/4500

It's not in the page title, it's not shown anywhere.

Doesn't need to be editable, perhaps, but should be at least visible.

PR #87, opened 2019-03-24
Updated matches backend

### LINKED ISSUE #85: On edit match page, "winning team" dropdown lists all teams (not even alphabetized) instead of just the ones in the match
e.g.
http://localhost:8080/hl/backend/rikki/heroeslounge/match/update/4500

Astraeus-:
Fixes #85 #86  
Removed the match round from update match. I don't think this needs to editable after a match has been created. 

Disabled updating of the play and schedule deadlines from the update match backend.

Renamed fields to be more clear, added more intuitive ordering and made create/update match consistent.

Display teams participating in a match and filtered the winner selection based on the these participating teams, instead of the old massive drop down.

### DIFF
--- a/plugins/rikki/heroeslounge/controllers/Match.php
+++ b/plugins/rikki/heroeslounge/controllers/Match.php
@@ -30,16 +30,11 @@ class Match extends Controller
         $form->addFields([
             'division@create' => [
                 'label' => 'Division',
-                'span' => 'left',
+                'span' => 'right',
                 'type' => 'dropdown',
                 'options' => Division::listDivisionsWithLongTitle(),
                 'showSearch' => false,
                 'emptyOption' => '-- No Division --'
-            ],
-            'round@create' => [
-                'label' => 'Round',
-                'span' => 'auto',
-                'type' => 'number'
             ]
             ]);
     }
--- a/plugins/rikki/heroeslounge/controllers/Match.php
+++ b/plugins/rikki/heroeslounge/controllers/Match.php
@@ -7,6 +7,7 @@ use Rikki\Heroeslounge\Models\Team;
 use Rikki\Heroeslounge\Models\Timeline;
 use Rikki\Heroeslounge\Models\Match as MatchModel;
 use Rikki\Heroeslounge\Models\Game;
+use Rikki\Heroeslounge\Models\Division;
 
 class Match extends Controller
 {
@@ -22,6 +23,25 @@ class Match extends Controller
         parent::__construct();
 
         BackendMenu::setContext('Rikki.Heroeslounge', 'manage-matches','manage-matches');
+    }
+
+    public static function formExtendFields($form)
+    {
+        $form->addFields([
+            'division@create' => [
+                'label' => 'Division',
+                'span' => 'left',
+                'type' => 'dropdown',
+                'options' => Division::listDivisionsWithLongTitle(),
+                'showSearch' => false,
+                'emptyOption' => '-- No Division --'
+            ],
+            'round@create' => [
+                'label' => 'Round',
+                'span' => 'auto',
+                'type' => 'number'
+            ]
+            ]);
     }
 
     public function onRelationButtonApprove()
--- a/plugins/rikki/heroeslounge/models/Division.php
+++ b/plugins/rikki/heroeslounge/models/Division.php
@@ -85,10 +85,21 @@ class Division extends Model
     {
         if ($this->season != null) {
             return $this->season->title . ' - ' . $this->title;
-        } else if ($this->playoff != null && $this->playoff->season != null) {
-            return $this->playoff->season->title . ' - ' . $this->playoff->title;
+        } else if ($this->playoff != null) {
+            return $this->playoff->longTitle . ' - ' . $this->title;
         }
         return $this->title;
+    }
+
+    public static function listDivisionsWithLongTitle()
+    {
+        $divisions = Division::all();
+        $list = [];
+
+        foreach ($divisions as $division) {
+            $list[$division->id] = $division->longTitle;
+        }
+        return $list;
     }
 
     public function getActiveTeamsAttribute()
--- a/plugins/rikki/heroeslounge/models/Division.php
+++ b/plugins/rikki/heroeslounge/models/Division.php
@@ -93,13 +93,7 @@ class Division extends Model
 
     public static function listDivisionsWithLongTitle()
     {
-        $divisions = Division::all();
-        $list = [];
-
-        foreach ($divisions as $division) {
-            $list[$division->id] = $division->longTitle;
-        }
-        return $list;
+        return Division::all()->keyBy('id')->map(function ($division) { return $division->longTitle; })->toArray();
     }
 
     public function getActiveTeamsAttribute()
--- a/plugins/rikki/heroeslounge/models/Match.php
+++ b/plugins/rikki/heroeslounge/models/Match.php
@@ -82,6 +82,11 @@ class Match extends Model
         ]
     ];
 
+    public function listWinnerOptions($fieldname, $value, $formData)
+    {
+        return [$formData["teams"][0]['id'] => $formData["teams"][0]['title'], $formData["teams"][1]['id'] => $formData["teams"][1]['title']];
+    }
+
     public function getCasterIds()
     {
         return $this->casters
--- a/plugins/rikki/heroeslounge/models/Match.php
+++ b/plugins/rikki/heroeslounge/models/Match.php
@@ -9,7 +9,6 @@ use Rikki\Heroeslounge\classes\hotfixes as hotfixes;
 use Rikki\Heroeslounge\Models\Season as Season;
 use Rikki\Heroeslounge\Models\Playoff;
 use Rikki\Heroeslounge\Models\Team;
-use Rikki\Heroeslounge\Models\Division;
 use Carbon\Carbon;
 use Log;
 /**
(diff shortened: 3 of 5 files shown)


rogusdev:
This looks great!  It does have side effects that were not at all expected to me -- that I further like,m but want to understand where they come from: Caster and Timeline have now moved to separate tabs, and the tab design has changed.

Old:
edit_match_old

New:
edit_match_new

How did those changes happen?  Very happy with them, just want to understand the system.

Astraeus-:
That's part of my backend refactor in 48540d1482f7c174c473b682673266a62d57591d specifically for matches: `plugins/rikki/heroeslounge/models/match/fields.yaml` and `plugins/rikki/heroeslounge/models/match/fields_on_create.yaml`

rogusdev:
Doh... I had not rebased my branch on latest upstream master... mb, thanks!

Astraeus-:
Also fixes #15 by adding a custom dropdown instead of purely the relation to the match model. @Fabian-Sommer when you deploy there are some divisions that belong to a playoff that doesn't exist. (playoff_id: 12). These need to be deleted or I need to account for divisions that used to belong to a playoff that does not exist anymore.

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -82,6 +83,32 @@ class Match extends Model
...
+        return [$formData["teams"][0]['id'] => $formData["teams"][0]['title'], $formData["teams"][1]['id'] => $formData["teams"][1]['title']];
+    }
+
+    public function listDivisionOptions($fieldname, $value, $formData)
+    {
+        $divisions = Division::all();
+        $list = [];
+
+        foreach ($divisions as $division) {
+            if (!is_null($division->season_id)) {
```
rogusdev: Should this logic live on the division model so that it can be reused elsewhere?  This kind of naming seems like it is not unique to this dropdown, potentially.

Astraeus-: I'm not sure if it's worth moving this to the division model. However I did find that there's already a method `getLongTitleAttribute()` on the division model. It would probably be worth updating that logic and using it to create the long division title for each division.

rogusdev: oh yeah you should definitely modify that method because it is almost exactly the same as this logic already, and it does not appear to be used anywhere...  (So changing it is totally safe and would be perfect here)

Astraeus-:
Added the logic to the division model instead of the match model. Slightly refactored the existing `longTitle` logic and used that instead of my own to populate the division drop down.

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -9,6 +9,7 @@
 use Rikki\Heroeslounge\Models\Season as Season;
 use Rikki\Heroeslounge\Models\Playoff;
 use Rikki\Heroeslounge\Models\Team;
+use Rikki\Heroeslounge\Models\Division as Division;
```
rogusdev: Is this `as` not redundant since that's what it is already? (Same for season above)

plugins/rikki/heroeslounge/models/match/fields.yaml
```diff
@@ -1,40 +1,43 @@
...
+            label: Schedule for
             span: auto
             type: datepicker
             mode: datetime
-            readOnly: true
-            tab: General
-        round:
-            label: 'Round'
-            span: left
-            type: number
```
Fabian-Sommer: Please allow round to be edited after creation - people sometimes make mistakes, this gives them the option to fix their mistake.

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -82,6 +83,16 @@ class Match extends Model
         ]
     ];
 
+    public function listWinnerOptions($fieldname, $value, $formData)
```
Fabian-Sommer: Ideally, this should be part of the controller since it is very specific for the use in backend. Not sure if it can be moved easily. If it stays in the model, please rename it to listParticipatingTeamsForBackend or something like this

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -82,6 +83,16 @@ class Match extends Model
...
+    {
+        return [$formData["teams"][0]['id'] => $formData["teams"][0]['title'], $formData["teams"][1]['id'] => $formData["teams"][1]['title']];
+    }
+
+    public function divisionsLongTitleDropdown()
+    {
+        return Division::listDivisionsWithLongTitle();
+    }
```
Fabian-Sommer: If at all possible, this should not be part of the model, but of the controller - This has nothing to do with the match at all!

Astraeus-: According to the [OctoberCMS documentation](https://octobercms.com/docs/backend/forms#form-fields) for form fields, the field definitions should be defined on the model class file, unless I'm misunderstanding their documentation.

For this specific case, because it's a form for creating a match, I'm quite certain that the accompanying `fields_on_create.YAML` file can only reference the match class to populate it's data.

plugins/rikki/heroeslounge/models/Match.php
Astraeus-: See below response, as it's somewhat similar.

plugins/rikki/heroeslounge/models/Match.php
Fabian-Sommer: Could you point me to the part of the docs that make you think so? Unless my understanding of MVC is entirely off this is not something that goes in the model.

Also, it should be static if October is really so inflexible that it HAS to be there

plugins/rikki/heroeslounge/models/Division.php
```diff
@@ -85,12 +85,23 @@ public function getLongTitleAttribute()
...
 
+    public static function listDivisionsWithLongTitle()
+    {
+        $divisions = Division::all();
+        $list = [];
+
+        foreach ($divisions as $division) {
+            $list[$division->id] = $division->longTitle;
+        }
+        return $list;
```
Fabian-Sommer: This can be done with a single line using the collection and map

Astraeus-: I attempted to use `map`, but I'm not quite sure how to get the output that I need. The following is the closest I got, but I require an array of key value pairs of `$divison->id => $division->longTitle` for attaching the correct division to a match.

```php
Division::all()->map(function ($division) {
    return [$division->id => $division->longTitle];
})>flatten()->toArray();

Output:
  0 => 'Season 4 - Division 1',
  1 => 'Season 4 - Division 2',
  2 => 'Season 4 - Division 3',
...

Desired output:
  1 => 'Season 4 - Division 1',
  2 => 'Season 4 - Division 2',
  3 => 'Season 4 - Division 3',
...
```

Advice how to accomplish this is appreciated.

plugins/rikki/heroeslounge/models/Match.php
Astraeus-: Moved to the match controller

Astraeus-:
Re-added the ability to update a match round. Also moved the `divisionlongtitle` dropdown off of the match model to the match controller model, making certain to only render it on the `create` form. I added the `round` field there as well, because I wasn't sure how else to maintain the field order.

I couldn't figure out how to reference the selected match model from the controller when updating, so I wasn't able to move `listParticipatingTeamsForBackend` to the controller.

plugins/rikki/heroeslounge/models/Division.php
Fabian-Sommer: I think the problem in your current solution is that you at first create small arrays and then use flatten which causes the keys to be enumerated. Maybe try 
`Division::all()->map(function ($division) {
    return $division->id => $division->longTitle;
})->toArray();`

rogusdev: Better languages can solve this problem the way you describe, but php is far from that good.  As written, this really is the best way in php.  Here is some discussion of the options:
https://stackoverflow.com/questions/11563119/how-to-convert-an-array-of-arrays-or-objects-to-an-associative-array

rogusdev: Oh although I dug a bit deeper and it turns out you can do it with array map if you jump through some extra hoops:
https://ryanwinchester.ca/posts/php-array-map-with-keys

Look at example #3 from the official docs:
https://www.php.net/manual/en/function.array-map.php

plugins/rikki/heroeslounge/models/Match.php
Fabian-Sommer: It is

plugins/rikki/heroeslounge/controllers/Match.php
```diff
@@ -24,6 +25,25 @@ public function __construct()
...
+                'showSearch' => false,
+                'emptyOption' => '-- No Division --'
+            ],
+            'round@create' => [
+                'label' => 'Round',
+                'span' => 'auto',
+                'type' => 'number'
+            ]
+            ]);
+    }
```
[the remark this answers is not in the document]
Astraeus-: Yeah that's how I understood how logic needs to be divided when using `yaml` forms from the OctoberCMS documentation. It made me feel like the best practice is to define most of the logic on the model class instead of the controller, which makes the current implementation look like a bit of a "hack" to me.

Alternatively we could no longer use `yaml` forms, but I think their ease of use and the required rewrite outweighs doing so.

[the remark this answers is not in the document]
Fabian-Sommer: I really disliked having a static method in Match that just returns the result of a static method in Division. With this, it convolutes a bit where the fields are defined, but I think that is ok considering the alternative.
However, I think that the Round can be moved back to the .yaml - at least I see no reason why it has to be here.

Side note: Defining fields for a specific view is something that I believe is good to have in the view or the controller, but not in the model

plugins/rikki/heroeslounge/models/Division.php
Fabian-Sommer: Did you have a look at https://octobercms.com/docs/services/collections
Which is what October uses in many places instead of a bare PHP array

Astraeus-: Figured it out by using `collection->keyBy()` before `collection->map()`.

Issue #35, opened 2018-10-29
Enhance playoff creation


Fabian-Sommer:
Currently, creating the matches of playoffs relies on hardcoded times for the bracket matches. There should be a backend interface to configure all times before creating the matches.

Fabian-Sommer:
Additionally, the timezone is ignored currently

PR #89, opened 2019-03-28
only send timezone at start of session, use everywhere

### LINKED ISSUE #65: Refactor how we use the users timezone
Currently, we do AJAX calls on every page that displays time to account for the users timezone. This induces high server load. It would be great if we could save the timezone across a session and use it directly.

rogusdev:
Resolves #65 

This ~~does not~~ DOES appear to work quite the way I want: even though it does successfully set the timezone on the session, and it seems to load it just fine, the check for "hasTimezone" seems cached or not updated somewhere because it sends the timezone repeatedly if you reload the page without clicking away and even then the behavior thereafter seems confusing -- sometimes it does, sometimes it doesn't.

EDIT: turns out ^ was because of all the AJAX calls, which are removed in #91 so all is fine, exactly per:

Additionally, tho probably in a separate PR, we will want to merge the partials back into the default.htm in a variety of components now that they have no reason to be loaded async (thereby saving more http requests)

### DIFF
--- a/plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
+++ b/plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
@@ -1,0 +1,52 @@
+<?php namespace Rikki\Heroeslounge\Classes\Helpers;
+
+use Auth;
+use Config;
+use Session;
+
+class TimezoneHelper
+{
+    private const TIMEZONE_KEY = 'timezone';
+
+    public static $defaultTimezone;
+
+    private static function defaultTimezone()
+    {
+        if (!self::$defaultTimezone)
+        {
+            self::$defaultTimezone = Config::get('app.timezone', 'Europe/Berlin');
+        }
+        return self::$defaultTimezone;
+    }
+
+    public static function getTimezone()
+    {
+        return Session::get(self::TIMEZONE_KEY, self::defaultTimezone());
+    }
+
+    public static function setTimezone()
+    {
+        if (isset($_POST[self::TIMEZONE_KEY])) {
+            $timezoneName = $_POST[self::TIMEZONE_KEY];
+        } else {
+            $timezoneName = self::defaultTimezone();
+        }
+        if (!in_array($timezoneName, timezone_identifiers_list())) {
+            $timezoneName = self::defaultTimezone();
+        }
+        Session::put(self::TIMEZONE_KEY, $timezoneName);
+
+        $user = Auth::getUser();
+        if ($user) {
+            if ($user->sloth->timezone == '') {
+                $user->sloth->timezone = TimezoneHelper::getTimezone();
+                $user->sloth->save();
+            }
+        }
+    }
+
+    public static function hasTimezone()
+    {
+        return Session::has(self::TIMEZONE_KEY);
+    }
+}
--- a/plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
+++ b/plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
@@ -6,15 +6,17 @@ use Session;
 
 class TimezoneHelper
 {
+    private const DEFAULT_TIMEZONE = 'UTC';
     private const TIMEZONE_KEY = 'timezone';
 
-    public static $defaultTimezone;
+    private static $defaultTimezone;
 
     private static function defaultTimezone()
     {
         if (!self::$defaultTimezone)
         {
-            self::$defaultTimezone = Config::get('app.timezone', 'Europe/Berlin');
+            self::$defaultTimezone =
+                Config::get('app.timezone', self::DEFAULT_TIMEZONE);
         }
         return self::$defaultTimezone;
     }
@@ -35,14 +37,6 @@ class TimezoneHelper
             $timezoneName = self::defaultTimezone();
         }
         Session::put(self::TIMEZONE_KEY, $timezoneName);
-
-        $user = Auth::getUser();
-        if ($user) {
-            if ($user->sloth->timezone == '') {
-                $user->sloth->timezone = TimezoneHelper::getTimezone();
-                $user->sloth->save();
-            }
-        }
     }
 
     public static function hasTimezone()
--- a/plugins/rikki/heroeslounge/components/ScheduleMatch.php
+++ b/plugins/rikki/heroeslounge/components/ScheduleMatch.php
@@ -10,6 +10,7 @@ use DateTime;
 use DateTimeZone;
 use Flash;
 use Carbon\Carbon;
+use Rikki\Heroeslounge\Classes\Helpers\TimezoneHelper;
 
 class ScheduleMatch extends ComponentBase
 {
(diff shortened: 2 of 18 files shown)


plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
```diff
@@ -0,0 +1,52 @@
...
+            $timezoneName = $_POST[self::TIMEZONE_KEY];
+        } else {
+            $timezoneName = self::defaultTimezone();
+        }
+        if (!in_array($timezoneName, timezone_identifiers_list())) {
+            $timezoneName = self::defaultTimezone();
+        }
+        Session::put(self::TIMEZONE_KEY, $timezoneName);
+
+        $user = Auth::getUser();
```
rogusdev: turns out, this never gets called because by the time you are logged in, the timezone is already set

plugins/rikki/heroeslounge/models/Playoff.php
```diff
@@ -47,10 +49,8 @@ public function getLongTitleAttribute()
     public function createMatches()
     {
         $playoff = $this;
-        $timezone = 'Europe/Berlin';
+        $timezone = TimezoneHelper::defaultTimezone();
         if ($this->type == 'playoffv1') {
-            //TODO configure dates and timezone from backend
```
rogusdev: should line 52 be `getTimezone()` now?

Astraeus-: I believe this requires a lot more work than just updating the timezone variable, as the automatic match creation is currently hard-coded, and manually changed by Zalz for every playoff creation

plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
Fabian-Sommer: In my experience, one can stay logged in on the page for an indefinite amount of time

plugins/rikki/heroeslounge/components/ScheduleMatch.php
```diff
@@ -69,21 +64,20 @@ public function onSaveDate()
         $match = Match::find(post('match_id'));
         if ($match) {
             $date = post('date');
-            $timezone = post('timezoneName');
+            $timezone = TimezoneHelper::getTimezone();
             if ($date != null) {
                 try {
                     $x = new DateTime($date, new DateTimeZone($timezone));
-                    $x->setTimezone(new DateTimeZone('Europe/Berlin'));
+                    $x->setTimezone(new DateTimeZone(TimezoneHelper::defaultTimezone()));
```
Fabian-Sommer: This needs to be the server timezone (UTC)

plugins/rikki/heroeslounge/components/ScheduleMatch.php
```diff
@@ -69,21 +64,20 @@ public function onSaveDate()
...
                     $match->wbp = $x->format('Y-m-d H:i:s');
                     if ($match->tbp != null && Carbon::parse($match->wbp) < Carbon::parse($match->tbp)) {
                         $match->save();
                         Flash::success('Match has been successfully scheduled for '.$date);
                     } else {
-                        $y = new DateTime($match->tbp, new DateTimeZone('Europe/Berlin'));
+                        $y = new DateTime($match->tbp, new DateTimeZone(TimezoneHelper::defaultTimezone()));
```
Fabian-Sommer: Also UTC

plugins/rikki/heroeslounge/models/Playoff.php
Fabian-Sommer: You can leave this function as is for now, its a different (large) issue. Last time I ran it, the $timezone parameter was ignored anyway.

plugins/rikki/heroeslounge/classes/helpers/TimezoneHelper.php
rogusdev: Yeah the login is permanent, but the session is per browser session, I think.

plugins/rikki/heroeslounge/models/Playoff.php
rogusdev: Should I leave the TODO?  Should I not change the timezone?

plugins/rikki/heroeslounge/components/ScheduleMatch.php
rogusdev: Yeah that's why the default timezone pulls from config/app

plugins/rikki/heroeslounge/components/ScheduleMatch.php
rogusdev: Same as above: it is, if that is properly configed, which you said it is.

plugins/rikki/loungeviews/components/Navigation.php
```diff
@@ -39,13 +39,19 @@ public function init()
...
         $this->divsseasons = Season::where('type', 2)->orderBy('created_at','desc')->get();
+
+        $this->user = Auth::getUser();
         if ($this->user != null) {
             $this->sloth = SlothModel::getFromUser($this->user);
+
+            if ($this->sloth->timezone == '') {
+                $this->sloth->timezone = TimezoneHelper::getTimezone();
+                $this->sloth->save();
+            }
```
rogusdev: I put this here because it seems the most likely to do the desired thing consistently

rogusdev:
I tested this and it appears to be working on the calendar page.  There was one confusing bit others might want to be aware of when testing: there is a timezone for the backend!
http://localhost:8080/hl/backend/backend/preferences
^your backend account has its own stored timezone that all times are converted into, separate from config/app.php's timezone.  The latter affects what's stored in the db, the former affects how those times are displayed in the UI.

So, I removed the WIP and it is ready for final reviewing, thanks.

plugins/rikki/heroeslounge/models/Playoff.php
Fabian-Sommer: Indeed, its part of #35

themes/HeroesLounge-Theme/layouts/blog.htm
```diff
@@ -12,30 +12,22 @@
 <?php
 use RainLab\Translate\Models\Message;
 use Rikki\Heroeslounge\Models\Sloth;
+use Rikki\Heroeslounge\Classes\Helpers\TimezoneHelper;
 
 function onStart() {
     $this['baseUrl'] = Config::get('app.url');
     Message::setContext(App::getLocale(), $this->page->url);
+    $this['hasTimezone'] = TimezoneHelper::hasTimezone();
```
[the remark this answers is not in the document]
rogusdev: I believe this is because of all the ajax.  Perhaps it is not setting the session key correctly in there somewhere.  Once the ajax has been removed, this can be revisited.  Right now, it is at least not worse than before, and should be better (less unnecessary calls).

rogusdev:
I intend to do the ajax merging in a separate PR, this one is complete for my tastes.

Fabian-Sommer:
The timezone should also get set on sites that do not rely on any other ajax, right (ruleset for example). Is it getting set correctly there? If not, the reason is probably not the other ajax calls.

rogusdev:
Good call!  I tested with http://localhost:8080/hl/guides/captains-guide and it only set once.  I have begun the PR for merging ajax calls, that should go up today or tomorrow, I expect.

rogusdev:
@Fabian-Sommer once this one is approved I will merge, rebase #91 and that one will be ready for review as well.  Changes are quite straightforward and a very nice cleanup

PR #92, opened 2019-03-29
Allow all team members to upload match replays

### LINKED ISSUE #10: Allow all team members to upload replays

Astraeus-:
Fixes #10 
Enables the `Matches` option for all team members in the left sidebar navigation. Also moved the views / manage options outside of the conditional to remove duplicates.

### DIFF
--- a/plugins/rikki/heroeslounge/components/managematches/default.htm
+++ b/plugins/rikki/heroeslounge/components/managematches/default.htm
@@ -4,7 +4,7 @@
 <div class="row mb-3">
     {% endif %} 
     {% if match.wbp == null %}
-    {% if __SELF__.user.sloth.is_captain or __SELF__.user.sloth.is_divs_captain %}
+    {% if __SELF__.user.sloth.isCaptain(__SELF__.team) %}
     {% component 'scheduleMatch' class="mb-3"  id=match.id %}
     {% endif %}
     {% else %} 
--- a/plugins/rikki/heroeslounge/components/managematches/default.htm
+++ b/plugins/rikki/heroeslounge/components/managematches/default.htm
@@ -3,8 +3,10 @@
 {% for match in __SELF__.matches %} {% if loop.index0 is even %}
 <div class="row mb-3">
     {% endif %} 
-    {% if match.wbp == null %} 
-    {% component 'scheduleMatch' class="mb-3"  id=match.id %} 
+    {% if match.wbp == null %}
+    {% if __SELF__.user.sloth.is_captain or __SELF__.user.sloth.is_divs_captain %}
+    {% component 'scheduleMatch' class="mb-3"  id=match.id %}
+    {% endif %}
     {% else %} 
     {%  component 'updateMatch' class="mb-3" id=match.id %}
     {% endif %} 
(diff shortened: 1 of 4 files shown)


plugins/rikki/loungeviews/components/sidenav/default.htm
```diff
@@ -28,32 +28,26 @@
...
+                                View
+                        </a>
                     </li>
                     <li>
                         <a href="{{ 'team/manageMatch' | page({slug: user.sloth.team.slug}) }}" title="Manage Matches">
                             <i class="fa fa-futbol-o" aria-hidden="true"></i>
                             Matches
                         </a>
                     </li>
+                {% if user.sloth.is_captain == true %}
```
rogusdev: No need for `== true`

plugins/rikki/loungeviews/components/sidenav/default.htm
```diff
@@ -71,32 +65,26 @@
...
+                            View
+                        </a>
                     </li>
                     <li>
                         <a href="{{ 'team/manageMatch' | page({slug: user.sloth.divs_team.slug}) }}" title="Manage Matches">
                             <i class="fa fa-futbol-o" aria-hidden="true"></i>
                             Matches
                         </a>
                     </li>
+                {% if user.sloth.is_divs_captain == true %}
```
rogusdev: Same

plugins/rikki/heroeslounge/components/managematches/default.htm
```diff
@@ -3,8 +3,10 @@
 {% for match in __SELF__.matches %} {% if loop.index0 is even %}
 <div class="row mb-3">
     {% endif %} 
-    {% if match.wbp == null %} 
-    {% component 'scheduleMatch' class="mb-3"  id=match.id %} 
+    {% if match.wbp == null %}
+    {% if __SELF__.user.sloth.is_captain or __SELF__.user.sloth.is_divs_captain %}
```
Fabian-Sommer: Please check that the sloth is also the captain of the team and not just captain of any team

rogusdev: Ha, good call.  There should be a method on the sloth model `isCaptainOf(team_id)`

Astraeus-: Would there be a preference for determining the captain through the team model vs sloth model? As they both have a method. I was personally leaning towards team model but I'm not sure.

Fabian-Sommer: Sloth model would be more natural for me, also possibly faster

plugins/rikki/heroeslounge/components/manageteam/default.htm
```diff
@@ -1,4 +1,4 @@
-{% if (__SELF__.user.sloth.is_captain == 1 and __SELF__.user.sloth.team_id == __SELF__.team.id) or (__SELF__.user.sloth.is_divs_captain == 1 and __SELF__.user.sloth.divs_team_id == __SELF__.team.id) %}
+{% if __SELF__.user.sloth.isCaptain(__SELF__.team) %}
```
rogusdev: I looked at the implementation of this method:
https://github.com/Fabian-Sommer/HeroesLounge/blob/master/plugins/rikki/heroeslounge/models/Sloth.php

And it does not have the correct logic at all.  We should update the logic there to match the logic you replaced here, as that looks correct.

Astraeus-: You're correct, I assumed the logic in there was fine, but it seems that it is not. Will change to properly reflect that the user is indeed the captain of the specified team

PR #94, opened 2019-03-31
api matches for today or date with timezone

### LINKED ISSUE #40: Matches today API endpoint
The current [matches today endpoint](https://heroeslounge.gg/api/v1/matches/today) is based on the current date on the server. This means that matches taking place in the US region after 15:00 PT are not included when making a request to the endpoint.
```
public function getTodaysMatches()
    {
        return json_encode(MatchModel::whereDate('wbp','=',date(Carbon::today()))->where('is_played',false)->get());
    }
```

I'm not entirely sure what the best solution would be, but probably along the lines of getting the games from the next day that are in NA before a certain time.

rogusdev:
Resolves #40 

http://localhost:8080/hl/api/v1/matches/today returns matches occurring today in default timezone (UTC, server time) -- same as before (well, since server timezone changed from Berlin to UTC, at any rate...)
http://localhost:8080/hl/api/v1/matches/today/Asia/Tokyo returns matches "today" in a given timezone (here Asia/Tokyo)
http://localhost:8080/hl/api/v1/matches/2019-04-03/Asia/Tokyo returns matches for a specific date in a given timezone (here Asia/Tokyo)
http://localhost:8080/hl/api/v1/matches/2019-04-03 returns matches for a specific date in default timezone (UTC)

http://localhost:8080/hl/api/v1/matches/2019-04-03/America/New_York is same idea, different timezone
http://localhost:8080/hl/api/v1/matches/today/UTC is still valid when someone wants to be explicit
http://localhost:8080/hl/api/v1/matches/2019-04-03/UTC is same idea, now explicitly UTC for a specific date

### DIFF
--- a/plugins/rikki/heroeslounge/routes.php
+++ b/plugins/rikki/heroeslounge/routes.php
@@ -31,9 +31,9 @@ Route::group(['prefix' => 'api/v1'], function () {
     Route::get('matches/withApprovedCastBetween/{startdate}/{enddate}','Rikki\Heroeslounge\Http\Match@withApprovedCastBetween');
     Route::get('matches/{id}/games','Rikki\Heroeslounge\Http\Match@games');
     Route::get('matches/{id}/replays','Rikki\Heroeslounge\Http\Match@replays');
-    Route::get('matches/today','Rikki\Heroeslounge\Http\Match@getTodaysMatches');
-    
-    
+    Route::get('matches/{date}/{tz1}/{tz2?}','Rikki\Heroeslounge\Http\Match@getTodaysMatches');
+    Route::get('matches/{date}','Rikki\Heroeslounge\Http\Match@getTodaysMatches');
+
     Route::resource('matches', 'Rikki\Heroeslounge\Http\Match');
     Route::get('matchesAll','Rikki\Heroeslounge\Http\Match@indexAll');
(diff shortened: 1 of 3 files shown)


rogusdev:
Where do you update API documentation?

plugins/rikki/heroeslounge/routes.php
```diff
@@ -31,9 +31,9 @@
     Route::get('matches/withApprovedCastBetween/{startdate}/{enddate}','Rikki\Heroeslounge\Http\Match@withApprovedCastBetween');
     Route::get('matches/{id}/games','Rikki\Heroeslounge\Http\Match@games');
     Route::get('matches/{id}/replays','Rikki\Heroeslounge\Http\Match@replays');
-    Route::get('matches/today','Rikki\Heroeslounge\Http\Match@getTodaysMatches');
```
Fabian-Sommer: I would like for this to stay around with day of server as default, otherwise anything that calls this will suddenly cease to work

rogusdev: It does, look 2 lines down, and note the handling of "today" in the method

rogusdev: It does, look 2 lines down, and note the handling of "today" in the method.  Also note my very first example above

Issue #72, opened 2019-03-05
Enable multiple twitch channels be attached to a match


Fabian-Sommer:
With Division S, many matches will be cast by various casters across multiple twitch channels. Currently, each match can only have one twitch channel attached. We want to change the one-to-many relationship to a many-to-many relationship.

This affects:
- Database model - we need a new match_channel table to hold the new relation. The channel_id column in the match table needs to be deleted - however, before this can happen the existing data in the column needs to be used to populate the new table (I can do that when setting it up on the server, just be aware not to delete the column before committing everything else)
- Models - both match and twitch channel will need to be updated to reflect the new relation type
- Backend (controllers) - the way attaching channels to matches needs to be redone. Currently it is a bit of custom code (this was done for reasons that are no longer relevant), it is ok to just add the relation to the match form and remove it from where it is now.
- Frontend - both calendar (upcoming matches) and match overview page.

This is an issue that takes you once through the entire system, so should be a nice first issue to work on.

PR #96, opened 2019-04-02
add regex for date param in matches

### LINKED ISSUE #95: Matches endpoint errors when requesting by specifc match id
The matches endpoint for a specific match is currently broken.
E.g: `https://heroeslounge.gg/api/v1/matches/3122` => `http://localhost:8080/hl/api/v1/matches/312`

I expect to get information for the specified match, but instead instead get an error. I assume this an unforeseen side-effect from the update to the `matches/today` endpoint by @rogusdev

rogusdev:
Resolves #95

### DIFF
--- a/plugins/rikki/heroeslounge/routes.php
+++ b/plugins/rikki/heroeslounge/routes.php
@@ -31,8 +31,8 @@ Route::group(['prefix' => 'api/v1'], function () {
     Route::get('matches/withApprovedCastBetween/{startdate}/{enddate}','Rikki\Heroeslounge\Http\Match@withApprovedCastBetween');
     Route::get('matches/{id}/games','Rikki\Heroeslounge\Http\Match@games');
     Route::get('matches/{id}/replays','Rikki\Heroeslounge\Http\Match@replays');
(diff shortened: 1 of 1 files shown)


Astraeus-:
Would it not have been more intuitive to define the endpoints as follows? 

`api/v1/matches/today/{tz1}/{tz2} ` defaulting to UTC 

`api/v1/matches/byDate/{date}/{tz1}/{tz2} ` defaulting to UTC

rogusdev:
> Would it not have been more intuitive to define the endpoints as follows?
> 
> `api/v1/matches/today/{tz1}/{tz2} ` defaulting to UTC
> 
> `api/v1/matches/byDate/{date}/{tz1}/{tz2} ` defaulting to UTC

I could go either way.  Easy enough to change.

I still don't know where API docs are, so I don't know if they've been updated to communicate these new options yet anyway.  As in, I imagine we could still change things.

Fabian-Sommer:
Currently, API docs are a link to the routes.php file (i.e. non-existent)

rogusdev:
@Astraeus- you have my vote to change the endpoints if you want, @Fabian-Sommer pref?

Fabian-Sommer:
No preference on my end

Astraeus-:
I'll update them when I do the other API changes related to #72

PR #97, opened 2019-04-02
default season for sloth and team stats plus participated seasons

### LINKED ISSUE #50: Default stats to NA season for NA players
When viewing user profiles for for an NA user, the season dropdown still defaults to EU Season 8 (e.g. "Displaying stats of [EU]Season 8"). We should default this to "[NA]Season 1".

rogusdev:
Resolves #50 

I would like extra testing from others because I'm not sure I have good examples in my local db to verify this is working -- specifically the active season handling, as well as verifying belongsToSeason refactor.

In addition / alternatively. I'd like some guidance on setting up my local env for testing this properly, if you could add details here:
https://github.com/Fabian-Sommer/HeroesLounge/wiki/Add-Active-Season-with-Played-Matches-for-Teams-and-Sloths

### DIFF
--- a/plugins/rikki/heroeslounge/models/Match.php
+++ b/plugins/rikki/heroeslounge/models/Match.php
@@ -103,6 +103,18 @@ class Match extends Model
         return $this->channels->count() > 0 ? $this->channels->first() : null;
     }
 
+    public function getSeasonAttribute()
+    {
+        if ($this->division != null && $this->division->season != null) {
+            return $this->division->season;
+        } elseif ($this->playoff != null && $this->playoff->season != null) {
+            return $this->playoff->season;
+        } elseif ($this->division != null && $this->division->playoff != null && $this->division->playoff->season != null) {
+            return $this->division->playoff->season;
+        }
+        return null;
+    }
+
     public function getCasterIds()
     {
         return $this->casters
@@ -143,18 +155,7 @@ class Match extends Model
 
     public function belongsToSeason($season)
     {
-        return (($this->division != null && $this->division->season != null && $this->division->season->id == $season->id)
-            || ($this->playoff != null && $this->playoff->season != null && $this->playoff->season->id == $season->id)
-            || ($this->division != null && $this->division->playoff != null && $this->division->playoff->season != null && $this->division->playoff->season->id == $season->id));
-    }
-
-    public function associatedSeasons()
-    {
-        return [
-            $this->division != null ? $this->division->season : null,
-            $this->playoff != null ? $this->playoff->season : null,
-            $this->division != null && $this->division->playoff != null ? $this->division->playoff->season : null
-        ];
+        return $this->season && $this->season->id == $season->id;
     }
 
     public function beforeUpdate()
--- a/plugins/rikki/heroeslounge/models/Match.php
+++ b/plugins/rikki/heroeslounge/models/Match.php
@@ -5,8 +5,8 @@ use Illuminate\Support\Facades\DB;
 use October\Rain\Database\Model;
 use Rikki\Heroeslounge\Models\Timeline;
 use October\Rain\Exception\SystemException;
-use Rikki\Heroeslounge\classes\hotfixes as hotfixes;
-use Rikki\Heroeslounge\Models\Season as Season;
+use Rikki\Heroeslounge\classes\hotfixes;
+use Rikki\Heroeslounge\Models\Season;
 use Rikki\Heroeslounge\Models\Playoff;
 use Rikki\Heroeslounge\Models\Team;
 use Carbon\Carbon;
@@ -148,6 +148,14 @@ class Match extends Model
             || ($this->division != null && $this->division->playoff != null && $this->division->playoff->season != null && $this->division->playoff->season->id == $season->id));
     }
 
+    public function associatedSeasons()
+    {
+        return [
+            $this->division != null ? $this->division->season : null,
+            $this->playoff != null ? $this->playoff->season : null,
+            $this->division != null && $this->division->playoff != null ? $this->division->playoff->season : null
+        ];
+    }
 
     public function beforeUpdate()
     {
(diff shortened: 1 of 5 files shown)


plugins/rikki/loungestatistics/components/SlothStatistics.php
```diff
@@ -24,19 +24,49 @@ public function componentDetails()
...
         $this->addJs('assets/js/datatables.min.js');
         $this->addJs('assets/js/loungestatistics.js');
         $this->addCss('assets/css/datatables.min.css');
+
         $this->sloth = Sloth::where('id', $this->property('sloth_id'))
                         ->with('gameParticipations', 'gameParticipations.team', 'gameParticipations.game', 'gameParticipations.game.map', 'gameParticipations.game.match', 'gameParticipations.hero', 'gameParticipations.game.teamOneFirstBan', 'gameParticipations.game.teamOneSecondBan', 'gameParticipations.game.teamTwoFirstBan', 'gameParticipations.game.teamTwoSecondBan')
                         ->first();
-        $this->allseasons = Season::where('reg_open', false)->get()->sortByDesc('created_at');
 
-        $this->calculateStats($this->allseasons->first());
```
rogusdev: did this line ever show anything?  It does not render to the partial like `onSeasonChange` below

plugins/rikki/loungestatistics/components/SlothStatistics.php
```diff
@@ -183,13 +213,15 @@ public function calculateStats($season)
...
+        if ($season_id == -1) {
+            $this->selectedSeason = null;
         } else {
-            $this->calculateStats(Season::find(input('season_id')));
+            $this->selectedSeason = Season::find($season_id);
         }
+        $this->calculateStats($this->selectedSeason);
         return [
-            '#slothstatistics' => $this->renderPartial('@stats')
+            '#teamstatistics' => $this->renderPartial('@stats')
```
Astraeus-: Should this not stay as sloth, since these aren't team statistics?

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -148,6 +148,14 @@ public function belongsToSeason($season)
             || ($this->division != null && $this->division->playoff != null && $this->division->playoff->season != null && $this->division->playoff->season->id == $season->id));
     }
 
+    public function associatedSeasons()
```
Astraeus-: I'm a bit confused, a match can only belong to one season at most.

- It belongs to a division, which can belong to a season.
- It belongs to a division, which can belong to a playoff, which can belong to a season.
- It belongs to a playoff, which can belong to a season.

I could be mistaken in the syntax, but it seems like you're returning an array of options, instead of the singular season it belongs to or null?

plugins/rikki/loungestatistics/components/SlothStatistics.php
rogusdev: Whoops, copy paste error, thanks!

plugins/rikki/heroeslounge/models/Match.php
rogusdev: I thought that too, but this is adapted from the method above it.

The point is that there are 3 different ways for a season to be associated with a match and I see no guarantees that they will all be the same.  If we just had one object to look at, like division and that had the season every time (if present), then that would be easy.  But the models do not appear to be that clean currently.

rogusdev: Said another way: if only one of these has a season, then it's fine.  And if they all have the same season, then it's still fine.

Astraeus-: It feels like you'd often be doing a lot of unneeded data retrieval, since a majority of our matches have a division season relation. I'm not sure how "expensive" it is to retrieve relational data, but I feel like a conditional would be more fitting to exit early.

rogusdev: True enough, if it is confirmed that the precedence order from the method above applies here as well, I can just return the first season encountered.

Fabian-Sommer: Do that, and also name the method getSeasonAttribute()
That way, you can just access it with $match->season.

Also please refactor Match::belongsToSeason to use this method

plugins/rikki/loungestatistics/components/SlothStatistics.php
```diff
@@ -24,19 +24,49 @@ public function componentDetails()
...
+            if ($a->created_at == $b->created_at) return 0;
+            return $a->created_at > $b->created_at ? 1 : -1;
+        });
+
+        if ($this->selectedSeason == null) {
+            $participatedSeasonsCount = count($this->participatedSeasons);
+            if ($participatedSeasonsCount > 0) {
+                $this->selectedSeason = $this->participatedSeasons[$participatedSeasonsCount - 1];
+            }
+        }
```
Fabian-Sommer: This whole section probably works, but would look better code-wise by using Collections - have a look at map, sort, filter and unique.

plugins/rikki/loungestatistics/components/TeamStatistics.php
```diff
@@ -24,18 +24,46 @@ public function componentDetails()
...
+            if ($a->created_at == $b->created_at) return 0;
+            return $a->created_at > $b->created_at ? 1 : -1;
+        });
+
+        if ($this->selectedSeason == null) {
+            $participatedSeasonsCount = count($this->participatedSeasons);
+            if ($participatedSeasonsCount > 0) {
+                $this->selectedSeason = $this->participatedSeasons[$participatedSeasonsCount - 1];
+            }
+        }
```
Fabian-Sommer: You could probably refactor most of this method and the one in slothstatistics into one - this duplicates a lot of code

plugins/rikki/loungestatistics/components/SlothStatistics.php
rogusdev: I was not clicking the "toggle" buttons, mb

plugins/rikki/loungestatistics/components/TeamStatistics.php
rogusdev: I did not consolidate them after my change to use collections because it would be more overhead to share code with their differences than to accept a bit of copy-paste

plugins/rikki/loungestatistics/components/SlothStatistics.php
rogusdev: unique does not work here, but I used collections

PR #98, opened 2019-04-02
Enhance API endpoints

Astraeus-:
Refactored the new match date endpoints.
Add the match channels endpoint to get the Twitch channels associated with a match.

Regarding the timezone code, do we want to add some validating for the input options? We don't do any checking of the data inputs or timezone validation. (That it exists as zone or region/city)

### DIFF
--- a/plugins/rikki/heroeslounge/http/Match.php
+++ b/plugins/rikki/heroeslounge/http/Match.php
@@ -83,7 +83,7 @@ class Match extends Controller
         return $retVal;
     }
 
-    private function formatTimezone($tz1, $tz2)
+    private static function formatTimezone($tz1, $tz2)
     {
         if ($tz2) {
             return $tz1 . '/' . $tz2;
@@ -92,31 +92,23 @@ class Match extends Controller
         }
     }
 
-    private function getMatchesForDate($start, $isPlayed)
+    private static function getDbDateMatches($start)
     {
         // Convert to default/server timezone for db query
         $start->setTime(0,0,0)->setTimezone(TimezoneHelper::DEFAULT_TIMEZONE);
         $end = $start->copy()->addDay();
 
-        if (!$isPlayed) {
-            return MatchModel::whereBetween('wbp', [$start, $end])->where('is_played',false)->get();
-        } else {
-            return MatchModel::whereBetween('wbp', [$start, $end])->get();
-        }
+        return MatchModel::whereBetween('wbp', [$start, $end])->where('is_played',false)->get();
     }
 
-    public function getTodaysMatches($tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
+    public function getMatchesForToday($tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
     {
-        $date = Carbon::today(Self::formatTimezone($tz1, $tz2));
-
-        return getMatchesByDate($date, false);
+        return self::getDbDateMatches(Carbon::today(self::formatTimezone($tz1, $tz2)));
     }
 
-    public function getMatchesByDate($date, $tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
+    public function getMatchesForDate($date, $tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
     {
-        $date = Carbon::createFromFormat('Y-m-d', $date, Self::formatTimezone($tz1, $tz2));
-
-        return getMatchesByDate($date);
+        return self::getDbDateMatches(Carbon::createFromFormat('Y-m-d', $date, self::formatTimezone($tz1, $tz2)));
     }
 
     public function caster($id)
--- a/plugins/rikki/heroeslounge/http/Match.php
+++ b/plugins/rikki/heroeslounge/http/Match.php
@@ -23,6 +23,11 @@ class Match extends Controller
         return MatchModel::all();
     }
 
+    public function channels($id)
+    {
+        return MatchModel::findOrFail($id)->channels;
+    }
+
     public function teams($id)
     {
         return MatchModel::findOrFail($id)->teams;
--- a/plugins/rikki/heroeslounge/http/Match.php
+++ b/plugins/rikki/heroeslounge/http/Match.php
@@ -78,7 +78,7 @@ class Match extends Controller
         return $retVal;
     }
 
-    public function getTodaysMatches($date, $tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
+    public function getTodaysMatches($tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
     {
         if ($tz2) {
             $timezone = $tz1 . '/' . $tz2;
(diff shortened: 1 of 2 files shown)


plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,36 @@ public function games($id)
...
+            $timezone = $tz1;
         }
 
+        $start = Carbon::createFromFormat('Y-m-d', $date, $timezone);
+
         // now convert back to default/server timezone for db query
         $start->setTime(0,0,0)->setTimezone(TimezoneHelper::DEFAULT_TIMEZONE);
         $end = $start->copy()->addDay();
-        return MatchModel::whereBetween('wbp', [$start, $end])->where('is_played',false)->get();
+        return MatchModel::whereBetween('wbp', [$start, $end])->get();
```
rogusdev: I used one method for both to avoid all this copy paste.  You can still have one method, just make it private and pass different args in from the calling methods that the api hooks into

Astraeus-: Did some refactoring, not entirely sure if I'm happy wit the outcome, but I think it's okay

plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,40 @@ public function games($id)
...
-        if ($date == 'today') {
-            $start = Carbon::today($timezone);
+    private function getMatchesForDate($start, $isPlayed)
+    {
+        // Convert to default/server timezone for db query
+        $start->setTime(0,0,0)->setTimezone(TimezoneHelper::DEFAULT_TIMEZONE);
+        $end = $start->copy()->addDay();
+
+        if (!$isPlayed) {
+            return MatchModel::whereBetween('wbp', [$start, $end])->where('is_played',false)->get();
```
Fabian-Sommer: This should probably use $isPlayed

plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,40 @@ public function games($id)
...
         } else {
-            $timezone = $tz1;
+            return $tz1;
         }
+    }
 
-        // get today in a given timezone
-        if ($date == 'today') {
-            $start = Carbon::today($timezone);
+    private function getMatchesForDate($start, $isPlayed)
```
Fabian-Sommer: I like getMatchesForDay better as name - this is also very close to getMatchesByDate

plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,40 @@ public function games($id)
...
-        $end = $start->copy()->addDay();
-        return MatchModel::whereBetween('wbp', [$start, $end])->where('is_played',false)->get();
+    public function getTodaysMatches($tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
+    {
+        $date = Carbon::today(Self::formatTimezone($tz1, $tz2));
+
+        return getMatchesByDate($date, false);
+    }
+
+    public function getMatchesByDate($date, $tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
```
Fabian-Sommer: I think this can actually be merged with getMatchesForDate above

plugins/rikki/heroeslounge/http/Match.php
rogusdev: I would call this `getDbDayMatches`

plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,40 @@ public function games($id)
...
+    {
+        $date = Carbon::createFromFormat('Y-m-d', $date, Self::formatTimezone($tz1, $tz2));
```
rogusdev: `self` should be lowercase

plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,40 @@ public function games($id)
...
         }
+    }
 
-        // now convert back to default/server timezone for db query
-        $start->setTime(0,0,0)->setTimezone(TimezoneHelper::DEFAULT_TIMEZONE);
-        $end = $start->copy()->addDay();
-        return MatchModel::whereBetween('wbp', [$start, $end])->where('is_played',false)->get();
+    public function getTodaysMatches($tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
+    {
+        $date = Carbon::today(Self::formatTimezone($tz1, $tz2));
```
rogusdev: `self` should be lowercase

plugins/rikki/heroeslounge/http/Match.php
rogusdev: Do you ever call `isPlayed` as true anywhere?

plugins/rikki/heroeslounge/routes.php
```diff
@@ -31,8 +31,9 @@
     Route::get('matches/withApprovedCastBetween/{startdate}/{enddate}','Rikki\Heroeslounge\Http\Match@withApprovedCastBetween');
     Route::get('matches/{id}/games','Rikki\Heroeslounge\Http\Match@games');
     Route::get('matches/{id}/replays','Rikki\Heroeslounge\Http\Match@replays');
-    Route::get('matches/{date}/{tz1}/{tz2?}','Rikki\Heroeslounge\Http\Match@getTodaysMatches')->where('date', 'today|\d{4}-\d{2}-\d{2}');
-    Route::get('matches/{date}','Rikki\Heroeslounge\Http\Match@getTodaysMatches')->where('date', 'today|\d{4}-\d{2}-\d{2}');
+    Route::get('matches/today/{tz1?}/{tz2?}','Rikki\Heroeslounge\Http\Match@getTodaysMatches');
+    Route::get('matches/byDate/{date}/{tz1?}/{tz2?}','Rikki\Heroeslounge\Http\Match@getMatchesByDate');
```
rogusdev: For consistency with `getTodaysMatches`, the other should be called `getByDateMatches`.

Alternatively, change to `getMatchesForToday` and leave other as `getMatchesByDate`

plugins/rikki/heroeslounge/http/Match.php
```diff
@@ -78,25 +83,40 @@ public function games($id)
...
+
+        return getMatchesByDate($date, false);
+    }
+
+    public function getMatchesByDate($date, $tz1 = TimezoneHelper::DEFAULT_TIMEZONE, $tz2 = "")
+    {
+        $date = Carbon::createFromFormat('Y-m-d', $date, Self::formatTimezone($tz1, $tz2));
+
+        return getMatchesByDate($date);
```
rogusdev: This is infinite recursion right now

rogusdev:
I implemented the changes I was suggested (plus some additional tweaks) for your consideration: https://github.com/rogusdev/HeroesLounge/commit/658d04d339bdff0a2e96ac8b42529be9ec2cdc58

Also happy to make a new PR or whatever.

Astraeus-:
> I implemented the changes I suggested (plus some additional tweaks) for your consideration: [rogusdev@658d04d](https://github.com/rogusdev/HeroesLounge/commit/658d04d339bdff0a2e96ac8b42529be9ec2cdc58)
> 
> Also happy to make a new PR or whatever.

I can agree with most of the changes you've made, but I'm not convinced on the name `dbday`. Also defaulting to only retrieve matches that aren't played makes matchesfordate unintuitive / useless since you'd only be able to use it for the future, which the name doesn't suggest. E. G upcomingMatchesForDate would be a more fitting name in that case.

rogusdev:
> defaulting to only retrieve matches that aren't played

The code you had written never actually used that, as I noted in my comments.  I see what you were aiming for now, I can add that.

What would prefer instead of `DbDay`?

rogusdev:
@Astraeus- I added one more set of tweaks then: https://github.com/rogusdev/HeroesLounge/commit/4ad1cceeb420a12c4882cb70ef735c0581721044

Astraeus-:
I think `dayMatches` is more exact, because it returns the matches for a given 24hr period from the given starting point. `dbDayMatches` comes across to me that no matter what I put in, I'm getting the closest matching 24hr period in server time.

For the rest it looks okay to me, however I still think we should consider checking that the date format is correct and/or the timezone can be formatted. Currently they're just meaningless http 500 responses when invalid.

rogusdev:
@Fabian-Sommer I updated the PR with my changes, @Astraeus- approves.  You onboard?

PR #103, opened 2019-04-14
whitespace cleanup

rogusdev:
Just a bunch of cleanup I encountered while doing other work that will be in subsequent PRs


plugins/rikki/loungeviews/components/playoffoverview/default.htm
```diff
@@ -148,7 +147,7 @@ <h6>Team Description:</h6>
                                 <div class="state">
                                     <time datetime="{{match['model'].wbp | date('m/d/Y', __SELF__.timezone)}}" dateformat="custom_medium" timeformat="none">{{match['model'].wbp | date('d M', __SELF__.timezone)}}</time>
                                     <div class="time"><time datetime="{{match['model'].wbp | date('m/d/Y', __SELF__.timezone)}}" dateformat="none" timeformat="short">{{match['model'].wbp | date('H:i', __SELF__.timezone)}}</time></div>
-                                    <div>Timezone: {{__SELF__.timezone}} {{__SELF__.timezoneOffset}}</div>
+                                    <div>({{__SELF__.timezoneOffset}})</div>
```
rogusdev: This is the one actual change in here because when looking at unscheduled playoff games the timezone is huge

plugins/rikki/loungeviews/components/seasonoverview/default.htm
```diff
@@ -1,10 +1,8 @@
...
+    {% if __SELF__.season.reg_open == true %}
+        {% component 'participationOverview' id=__SELF__.season.id %}
+    {% elseif  __SELF__.season.divisions %}
+    {% endif %} 
 {% else %}
-
-
-{% endif %}
\ No newline at end of file
+No such season exists!
```
rogusdev: I added a tiny bit of extra info for bad URLs

plugins/rikki/loungeviews/components/playoffoverview/default.htm
Fabian-Sommer: I think for this page it would be even better to just show the timezone in a single place at the top or bottom instead of repeating the same information for every scheduled match.

rogusdev: Excellent point.  I put it here because time and date might not actually be displayed anywhere on the page but I think your suggestion makes the page much, much cleaner and a timezone reminder without times will still be fine.

This has been updated now.

plugins/rikki/loungeviews/components/playoffoverview/default.htm
```diff
@@ -100,8 +100,8 @@ <h6>Team Description:</h6>
                     {% partial '@divisionView' div=division %}
                 </div>
                 {% endfor %}
-                
 
+                <div>All times in {{__SELF__.timezone}} {{__SELF__.timezoneOffset}}</div>
```
rogusdev: Here the timezone is moved up to the top of the tab rather than being with every displayed time

PR #104, opened 2019-04-14
timezone tracking refactor into component

rogusdev:
Also refactor the page start for all the layouts that reuse the same head with timezone setting


rogusdev:
@Fabian-Sommer is there anywhere in the rendering call stack where I could hook in early such that if TimezzoneHelper::hasTimezone() is false, we redirect users to a nearly blank page that just has the client side timezone setting javascript, without any layout or anything, such that it would load quickly?  I realized in testing this that on some pages, the refresh is very perceptible.

Fabian-Sommer:
Something that I have experimented with is that on the different layouts, we could maybe do a 
if hasTimezone
   //normal code
else
  //almost empty page
endif

rogusdev:
I would greatly prefer to do it in one place, early in the page loading process as this crosses all the layouts, really.

Btw I did notice that timezone is not (in prod -- before this pr fixes it, I hope) getting set on mobile views of the site.

rogusdev:
I would like to merge this PR as-is because it is a significant improvement over the state before.

I do have a separate commit I'd like to discuss after this is merged that redirects users to a small fast page:
https://github.com/rogusdev/HeroesLounge/commit/ee169008464b76ef2ada2f6fc9a475a7f7a6ce18

Which is not quite what I want, because ideally I'd just change what the page was going to render, then reload and suddenly it is where we wanted to be with no url change or anything.  But this is the general idea for what I'd like to explore -- after this PR is merged, I do not want to block these changes.

I am genuinely concerned that people are scheduling things on their phones and always getting UTC there, because until this PR is merged, there is no timezone setting for phones!

Astraeus-:
The time does update on the website on my phone according to the timezone I'm in, but I'm not sure what browser you're using.

rogusdev:
Hmm, I don't see any other component named in that format, to my mind, I actually chose this name because it seemed the most consistent with the other components

PR #105, opened 2019-04-14
spoilers component on relevant pages

### LINKED ISSUE #62: Add spoiler filter
Some users want to see playoff brackets and schedules without seeing results, especially for Div S. We could provide a switch similar to how HGC did it.

rogusdev:
Resolves #62

spoilers_example


rogusdev:
Looking for feedback on this direction

Astraeus-:
I tested this locally and it seems to not be working for playoff divisions. I tried Season 7 Cup Group A, where toggling the spoiler option didn't make the scores visible. Also if possible it would be prettier to have the "spoiler blocks" be the same color as the background instead of the gray, they stand out quite a bit right now.

rogusdev:
The standout was a conscious choice, to indicate that something is hidden.  If they are just blank, users will likely be confused -- by default spoilers are not shown

rogusdev:
Can you give me a link to the specific playoff division where you had problems?  I thought I tested this everywhere-- you can see the links I primarily used linked from the issue ticket

Astraeus-:
I tried locally on `/season-7/playoff/Cup` then selecting Group A. It worked just fine for the tournament bracket.

rogusdev:
The match scores are hidden, tho I leave the standings alone since I'm not sure if that's necessary to hide or not.
spoilers_example_2

Astraeus-:
My problem is that I can't get them to show up. I've tried toggling the spoiler multiple times, refreshing the page, but no luck.

image

rogusdev:
Can you check your browser console and tell me of any errors you see?

Astraeus-:
Not getting any errors, on Firefox Version 66.0.3 all the other division tables work just fine, as well as playoff bracket. Just the groups that aren't updating.

rogusdev:
That is extremely odd.  Could you do some digging on your end to figure it out?  I cannot repro this problem. :(

Astraeus-:
I think I narrows it down to be related to the below snippet:
https://github.com/Fabian-Sommer/HeroesLounge/blob/f05fa2ab167e38d6cd2477b02c3d8ee59b52d678/themes/HeroesLounge-Theme/pages/playoff/view.htm#L14-L21


This updates the spoilers for the bracket stage:
`<div class="result win">2</div>` 
whereas the division match listings are badges and don't get updated:
`<span class="f100 spoiler badge badge-success">2</span>`

rogusdev:
Ahhhh got it now, thank you so much for testing it so thoroughly!  Fixed now.

Astraeus-:
I saw that you made changes for the season playoffs. I'm not sure if currently the normal playoffs support groups, because these would break in the future. @Fabian-Sommer would know if that's the case though.

rogusdev:
I added it for the regular playoffs as well to be safe, thanks

plugins/rikki/heroeslounge/assets/css/heroeslounge.css
```diff
@@ -401,6 +401,10 @@
 .match.format-mini>.record>.opponent>.result.loss {
     color: #e61a1a
 }
+.spoiler {
+    color: gray !important;
```
Astraeus-: Is there a specific reason for this to need `!important`?

rogusdev: the colors for the same elements are the ones just above that have much more specific selectors, which means that without !important the spoiler color would not be used.

It might not be required for the background color, but I figured I'd stick with it just in case.

plugins/rikki/loungeviews/components/spoilerstoggle/default.htm
```diff
@@ -0,0 +1,32 @@
+<div>
+    <span>Show Spoilers</span>
```
Fabian-Sommer: I think this should be about the same size vertically as the switch

rogusdev:
Cool, I was unsure what the scope should be so I'll make those changes.

rogusdev:
All of that is done now @Fabian-Sommer -- note that I did the toggle resize last so the first 4 screens are of the old toggle:

hlgg_division_spoilers
hlgg_division_spoilers_2
hlgg_playoff_spoilers
hlgg_playoff_spoilers_2
hlgg_spoilers_toggle_resized

plugins/rikki/loungeviews/components/spoilerstoggle/default.htm
rogusdev: done

plugins/rikki/loungeviews/components/playoffoverview/default.htm
```diff
@@ -117,29 +117,25 @@ <h6>Team Description:</h6>
                                 <div class="record">
                                     <div class="opponent opponent-1">
                                         {%if match['model'].teams[0] %}
-                                        {%if match['model'].teams[0].logo is not null%}<div class="logo"><img src="{{match['model'].teams[0].logo.path | resize(64,64)}}" style="position: relative; padding-top:0px;"title="{{match['model'].teams[0].title}}"></div>{%endif%}
-                                        <div class="name">{{match['model'].teams[0].title}}</div>
-                                        {%else%}
-                                        <div class="name disabled">TBD</div>
-                                        {%endif%}
+                                        <div class="logo spoiler">{%if match['model'].teams[0].logo is not null%}<img src="{{match['model'].teams[0].logo.path | resize(64,64)}}" style="position: relative; padding-top:0px;"title="{{match['model'].teams[0].title}}">{%endif%}</div>
```
rogusdev: The if statement needs to be inside the div not outside, otherwise any team with no logo does not line up properly and also is revealed in the later playoff rounds due to the discrepancy

Astraeus-:
I don't know if it's possible or how much work it would be to ignore hiding match results for teams playing BYE in the playoff bracket. 

For example in your second screenshot you already know round 2 will have gaming in the rhythm vs fortytwoesports as the two teams.

rogusdev:
I gotchu covered:
hlgg_playoff_spoilers_bye

plugins/rikki/loungeviews/components/playoffoverview/default.htm
```diff
@@ -117,29 +117,25 @@ <h6>Team Description:</h6>
...
+                                        <div class="name spoiler">{{match['model'].teams[0].title}}</div>{%else%}<div class="name disabled">TBD</div>{%endif%}
```
Fabian-Sommer: I would like to maintain the linebreaks here (bevore the else) so that it is clear which if it refers to

plugins/rikki/loungeviews/components/playoffoverview/default.htm
```diff
@@ -117,29 +117,25 @@ <h6>Team Description:</h6>
...
                                             {% endif %}
                                         {% endif %}
                                     </div>
                                     <div class="opponent opponent-2">
                                         {%if match['model'].teams[1] %}
-                                        {%if match['model'].teams[1].logo is not null%}<div class="logo"><img src="{{match['model'].teams[1].logo.path | resize(64,64)}}" title="{{match['model'].teams[1].title}}"></div>{%endif%}
-                                        <div class="name">{{match['model'].teams[1].title}}</div>{%else%}<div class="name disabled">TBD</div>
-                                        {%endif%}
+                                        <div class="logo spoiler">{%if match['model'].teams[1].logo is not null%}<img src="{{match['model'].teams[1].logo.path | resize(64,64)}}" title="{{match['model'].teams[1].title}}">{%endif%}</div>
+                                        <div class="name spoiler">{{match['model'].teams[1].title}}</div>{%else%}<div class="name disabled">TBD</div>{%endif%}
```
Fabian-Sommer: same here

PR #112, opened 2019-05-14
hide team names after byes in the losers bracket

rogusdev:
Before:
losers_bracket_broken

After:
losers_bracket_fixed


plugins/rikki/loungeviews/components/PlayoffOverview.php
```diff
@@ -136,13 +141,19 @@ public function onRender()
         $this->timeFormat = TimezoneHelper::getTimeFormatString();
     }
 
-    //gets offsets in rem for a playoff_position $pp
-    public function getOffsetsForMatch($pp)
+    private function getBracketName($dec_position) {
```
Fabian-Sommer: For consistency, lets put the bracket in its own line

rogusdev: updated

PR #125, opened 2019-06-15
Backend update captain

Astraeus-:
Started on the backend team captain management and captains of disbanding teams. I'm not entirely happy with how the sync is done with Discord when a team disbands, but it works.

Out of the box the team captain pivot data updates fine, but to incorporate the Discord captain role management I'm trying to use `onRelationManagePivotUpdate()` in the team controller. However when I add this listener it requires me to implement the entire pivot data update process and when changing the captain. a refresh is needed before the form view displays the made changes.

I'm not sure this is the right way to go about things, and any pointers would be great.

### DIFF
--- a/plugins/rikki/heroeslounge/controllers/team/config_relation.yaml
+++ b/plugins/rikki/heroeslounge/controllers/team/config_relation.yaml
@@ -59,15 +59,11 @@ sloths:
         list:
             columns:
                 title:
-                    label: title
+                    label: User
                     searchable: true
                 pivot[is_captain]:
-                    label: Captain
+                    label: Is Captain
                     type: switch
-                teams:
-                    label: Teams
-                    relation: teams
-                    select: title
         toolbarButtons: add|remove
     manage:
         showSearch: true
@@ -85,3 +81,4 @@ sloths:
                     type: checkbox
                     default: false
                     span: right
+                    
--- a/plugins/rikki/heroeslounge/controllers/team/config_relation.yaml
+++ b/plugins/rikki/heroeslounge/controllers/team/config_relation.yaml
@@ -81,4 +81,3 @@ sloths:
                     type: checkbox
                     default: false
                     span: right
-                    
--- a/plugins/rikki/heroeslounge/models/SlothTeamPivot.php
+++ b/plugins/rikki/heroeslounge/models/SlothTeamPivot.php
@@ -1,0 +1,45 @@
+<?php namespace Rikki\Heroeslounge\Models;
+
+use October\Rain\Database\Pivot;
+use Rikki\Heroeslounge\Models\Sloth;
+use Rikki\Heroeslounge\Models\Team;
+
+/**
+ * Sloth-Team Pivot Model
+ */
+class SlothTeamPivot extends Pivot
+{
+    
+    /*
+     * Disable timestamps by default.
+     * Remove this line if timestamps are defined in the database table.
+     */
+    public $timestamps = false;
+
+    /**
+     * @var string The database table used by the model.
+     */
+    public $table = 'rikki_heroeslounge_sloth_team';
+
+    public function afterDelete()
+    {
+        /*
+            I was hoping this would fire when a team gets disbanded, but it doesn't.
+        */
+    }
+
+    public function afterSave()
+    {
+        if ($this->isDirty('is_captain')) {
+            $team = Team::find($this->team_id);
+            $sloth = Sloth::find($this->sloth_id);
+
+            if ($sloth->isCaptainOfTeam($team)) {
+                $sloth->addDiscordCaptainRole();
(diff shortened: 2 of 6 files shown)


Fabian-Sommer:
Is it possible to put onRelationManagePivotUpdate() in the model instead? Then we could also remove the discord integration from the other places and have it in a single spot.

Astraeus-:
I wasn't able to use `onRelationManagePivotUpdate()` inside the model, because I believe it's a relation controller specific method. Instead I decided to define a model definition for the pivot data and that seems to work quite well on the backend. (I still need to check if the front-end captain switching works)

I updated the team disbanding to actually delete the pivot relation data, because sloths aren't part of the disbanding team anymore. I was hoping that this would cause the `afterDelete()` to fire in my `SlothTeamPivot` model, but that doesn't appear to be the case.

Until then, this is the best way I can come up with for handling the Discord captain integration. Any suggestions would be welcome.

plugins/rikki/heroeslounge/models/SlothTeamPivot.php
```diff
@@ -0,0 +1,50 @@
...
+                if ($isCaptain != null) {
+                    $noLongerCaptain = true;
+                }
+            }
+
+            if ($this->is_captain) {
+                $sloth->addDiscordCaptainRole();
+            } else if (!$this->is_captain && $noLongerCaptain) {
+                $sloth->removeDiscordCaptainRole();
+            }
```
Fabian-Sommer: I dislike this part for several reasons. First, it is quite long and for the most time the purpose of this code remains unclear - only in the last few lines is it revealed that it is all for the Discord integration. Maybe a comment or refactoring into a method could help with that.
Second: if ($this->is_captain) holds, $noLongerCaptain is not needed, even though half of the code deals with calculating it. Please only calculate it in the else if block in what is currently line 45 (and make an else out of the else if). While you are at it, rename $noLongerCaptain. It does not really convey the correct meaning, what it actually captures is whether there are other teams that the sloth is a captain of. Also, $isCaptain actually holds a team which is not what the name conveys. At least one of the two variables can be deleted.
Are lines 30/31/32/40 actually needed? From what I can see the else block should be fine even when there is only one team.

plugins/rikki/heroeslounge/controllers/team/config_relation.yaml
```diff
@@ -84,3 +80,4 @@ sloths:
                     type: checkbox
                     default: false
                     span: right
+
```
Fabian-Sommer: Please, clean this up

plugins/rikki/heroeslounge/models/SlothTeamPivot.php
```diff
@@ -0,0 +1,51 @@
...
+
+    public function isCaptainOtherTeam($sloth, $updatedTeamId) {
+        $teamCaptaining = $sloth->teams->first(function($index, $team) {
+            return $team->id != $updatedTeamId && $team->pivot->is_captain;
+        });
+
+        if ($teamCaptaining != null) {
+            return true
+        }
+        return false;
```
Fabian-Sommer: return true is missing a ;
In any case, I think this can be boiled down to return $sloth->... != null.
Probably better to use filter instead of first and then count() >= 1 instead of comparing with null, this should make the code even more readable.
I think the method itself should be named isCaptainOfOtherTeam or leadsOtherTeam

plugins/rikki/heroeslounge/models/SlothTeamPivot.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public $table = 'rikki_heroeslounge_sloth_team';
+
+    public function afterSave()
+    {
+        if ($this->isDirty('is_captain')) {
+            $sloth = Sloth::find($this->sloth_id);            
+
+            if ($this->is_captain) {
+                $sloth->addDiscordCaptainRole();
```
Fabian-Sommer: This causes unnecessary calls to the Discord API but this should be ok

plugins/rikki/heroeslounge/models/SlothTeamPivot.php
```diff
@@ -0,0 +1,51 @@
...
+            } else {
+                // We need to check if the sloth is not captain of another team before syncing Discord.
```
Fabian-Sommer: I think with improved readability the comment is no longer necessary

plugins/rikki/heroeslounge/models/SlothTeamPivot.php
Astraeus-: Yeah I debated doing a check if the user was already a captain, but I also remembered that we've done captain purging in the past, so the website is not always 100% in sync with Discord in that regard...

PR #127, opened 2019-06-19
Use the playoff title for matches in calendar

### LINKED ISSUE #124: Enhance Calendar
In the UpcomingMatches component, when a match has a division, the division title is displayed. This should be the playoff title if the match belongs to a playoff or the division belongs to a playoff

Astraeus-:
Uses the playoff title for upcoming matches instead of the division title for matches that belong to a playoff.

Fixes #124

### DIFF
--- a/plugins/rikki/heroeslounge/components/upcomingmatches/default.htm
+++ b/plugins/rikki/heroeslounge/components/upcomingmatches/default.htm
@@ -21,7 +21,15 @@
                         [{{match.teams[0].region.title}}]
                         {% endif %}
                     </span>
-                    <span title="{{match.division.title}}" class="hideOverflow">{{match.division.title}}</span>
+                    <span class="hideOverflow">
+                        {% if match.division.playoff is not null %}
+                        {{match.division.playoff.title}}
+                        {% elseif match.playoff is not null %}
+                        {{match.playoff.title}}
+                        {% else %}
+                        {{match.division.title}}
+                        {% endif %}
+                    </span>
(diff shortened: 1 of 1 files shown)


plugins/rikki/heroeslounge/components/upcomingmatches/default.htm
```diff
@@ -21,7 +21,15 @@
...
                     </span>
-                    <span title="{{match.division.title}}" class="hideOverflow">{{match.division.title}}</span>
+                    <span class="hideOverflow">
+                        {% if match.division.playoff is not null %}
+                        {{match.division.playoff.title}}
+                        {% elseif match.playoff is not null %}
+                        {{match.playoff.title}}
+                        {% else %}
+                        {{match.division.title}}
+                        {% endif %}
```
rogusdev: This seems like logic we will want to reuse elsewhere so should it be added on the match model?

Astraeus-: I wasn't entirely sure, but I don't think it's warranted. From the top of my head I can't think of another place where we need to be so selective with displaying match info.

rogusdev: I think I already did this somewhere actually, I just can't remember where

Fabian-Sommer: There is something similar in ViewMatch, but it is not quite the same

PR #133, opened 2019-06-22
Combine left sidebar with top navbar

Astraeus-:
Removes the left sidebar navigation and all references to it that I could find. The logic that was there is now on the top navbar.

Here's an example of the new look:
image

Any feedback / suggestions for additional changes are welcome.


plugins/rikki/loungeviews/components/Navigation.php
```diff
@@ -38,6 +39,10 @@ public function componentDetails()
     public $past_amateurseasons = null;
     public $divsseasons = null;
 
+
+    public $hasMatches = false;
+    public $seasons = null;
+
```
Fabian-Sommer: Are these required?

Astraeus-: Missed these when I copied over the old sidenav info. they didn't appear to be needed, so removed them.

Issue #64, opened 2019-02-15
Enhance API


Fabian-Sommer:
Requested endpoints:

- Division recent results: /division/{id}/recentresults => returns an array containing recent results (last day?). Entries need team names, logos and the score

- Division standings: /division/{id}/standings => returns an array containing the standings table. Entries need position, team name, team logo, wins, losses, map wins.

- Division standings team: /division/{division}/standings/{team} => returns the standings info. Entry needs position, team name, team logo, wins, losses, map wins.

- Division hero statistics: /division/{id}/herostats => returns an array containing hero participation statistics. Entries need number of picks, number of bans, number of wins.

- Role data: /roles/{id} => returns the role name (tank, bruiser, etc)

- Sloth hero data: /sloth/{id}/herostats => returns an array containing heroes played statistics. Entries need hero name, times played, wins.

Most of the data is already being calculated somewhere (except sloth hero data), but especially for stats and recent results this only happens in the component that renders the data. This needs to be refactored to happen in the model, so both component and API can access it.

rogusdev:
Ha. Free for all then?  Or any particular ways desired for enhancement?  Even just a link to a chat in discord to give context could spruce this one up to be more easily actionable.

Fabian-Sommer:
I edited the original post with more concrete requests

PR #139, opened 2019-08-10
Enhance api

Astraeus-:
Started on some of the things from issue #64. Wanted to know if you think I'm on the right track to continue. 

For the refactoring I was a bit unsure how to go about it. The current division implementation that I have now steals a lot from the teamstatistics component in the loungestatistics plugin, taking only the information out that's needed. (I kept the overall hero information, but that can be removed if you think it's not needed there).

For the slothstatistics I was going to go about it in a similar fashion, copying the bits I need to fulfill the listed requirement. Do you think that's the correct way to do this, or would you want all the in-game statistical information available as well? If so I would refactor all of it on to the sloth model as you had requested.


plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -0,0 +1,166 @@
...
+                // Check if we are the game winner.
+                $winner = ($game->winner_id == $team->id);
+
+                if ($game->map && $game->replay) {
+                    if ($team->title == $game->getSecondPickTeam()) {
+                        $mapArray[$game->map->title]['picks_by']++;
+                    } else {
+                        $mapArray[$game->map->title]['picks_vs']++;
+                    }
+                    if ($winner) {
```
Fabian-Sommer: You can omit $winner here and replace it with the condition above

plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -0,0 +1,166 @@
...
+
+        if ($type == "sloth" || $type == "slothAll") {
+            foreach ($rawData as $gP) {
+                if ($gP->hero == null) {
+                    continue;
+                }
+                if ($season == null || ($gP->game != null && $gP->game->match != null && $gP->game->match->belongsToSeason($season))) {
+                    $game = $gP->game;
+                    $team = $gP->team;
+                    if ($game == null or $team == null) {
```
Fabian-Sommer: These should just be checked above

Fabian-Sommer: (alongside hero != null)

plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -0,0 +1,166 @@
...
+                        continue;
+                    }
+                    
+                    // Find out team information.
+                    $winner = ($game->winner_id == $team->id);
+                    $tO = ($game->team_one_id == $team->id);
```
Fabian-Sommer: $tO is not a good name

plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -0,0 +1,166 @@
...
+                    $game->getTeamTwoBans()->each( function ($item) use ($tO, &$heroesArray) {
+                        if ($tO) {
+                            $heroesArray[$item->title]['bans_against_team']++;
+                        } else {
+                            $heroesArray[$item->title]['bans_by_team']++;
+                        }
+                    });
+    
+                    $heroesArray[$gP->hero->title]['picks']++;
+                    if ($winner) {
```
Fabian-Sommer: replace $winner by condition above

plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -0,0 +1,166 @@
...
+        Current supported types are:
+            - division -> Division api endpoint.
+            - sloth -> sloth hero statistics endpoint.
+            - slothAll -> sloth statistics component.
+
+        Our current types of rawData are:
+        - Matches
+        - gameParticipations        
+    */
+    public static function calculateHeroStatistics($type, $rawData, $season)
```
Fabian-Sommer: Please make sure that these parameters always have the same type - $rawData can be either matches or gameParticipations. In any case, please write in the comment which $type supports which $rawData (or which parameters).

plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -0,0 +1,166 @@
...
+                        if ($game->winner_id == $gP->team->id) {
+                            $heroesArray[$gP->hero->title]['wins']++;
+                        }                    
+                    }
+                }
+            }
+        }
+
+        return new Collection($heroesArray);
+    }
```
Fabian-Sommer: This function is too long overall - but both the Division part and the Sloth part can be split off into their own functions

PR #144, opened 2019-09-04
Start on Heroes Profile incorporation

Astraeus-:
Started work on adding Heroes Profile data to the website. Still unsure about some of the best ways to go about things. Also some things may need some additional refactoring.

Heroes Profile doesn't have an activity check to see if you're active enough in a certain gameMode, I just stuck to our current weighting system and don't think we can take into account how active people are.

Heroes Profile URL's are a bit of a pain to link to, since they require not only their unique identifier, but also the name portion of a battletag and the region (As similar to Blizzard replay regions). I'm not much of a Regex magician and we'll need to either create a dictionary to translate our regions to blizzard regions or migrate our own region definitions.

I also need to make the Heroes Profile MMR data available on the backend to view, and decide how to handle team MMR calculations(Split these up similarly to how we're going to do for Sloths)?

Pre-Deployment:
- We'll need to create an Auth file that will hold our Heroes Profile API Token. Not entirely sure what the best place for this would be.
- Run the migration (I just added the new data options we'll have, so we can have both HotsLogs and HeroesProfile in the future if we want them)


Astraeus-:
I've applied all the feedback that you sent to me.

The auth file for Heroes Profile expects to be in `Rikki\Heroeslounge\classes\MMR`. I don't know if you want to have unique auth files for different tokens, or share the file in `Rikki\Heroeslounge\classes\Discord`. I've made the name more distinctive, so it's up to you.

I haven't updated the schedule in `plugin.php` to retrieve the Heroes Profile data as I didn't want to have them potentially run before migrations.

plugins/rikki/heroeslounge/models/Team.php
```diff
@@ -193,7 +193,7 @@ public function getCaptainAttribute()
     public function getSlothratingAttribute()
     {       
         $slothsMmr = $this->sloths->map(function ($sloth) {
-            return $sloth->mmr;
+            return $sloth->heroesprofile_;
```
Fabian-Sommer: This misses mmr at the end (also a few lines below)

plugins/rikki/loungeviews/components/Profile.php
```diff
@@ -20,8 +20,8 @@ public function componentDetails()
     public $profile = null;
     public $sloth = null;
     public $user = null;
-    public $battleTagFormmattted = null;
-    public $hp_region = null;
+    public $battleTagFormatted = null;
+    public $region = null;
```
Fabian-Sommer: $region is not an ideal name since we already have a region attached here (and it is exactly NOT the one that is variable holds) - maybe heroesprofile_region_id

PR #150, opened 2019-09-18
Try to attach sub by heroesprofile_id first

### LINKED ISSUE #148: Wrong Heroes Lounge account associated with game
In this [match](https://heroeslounge.gg/match/view/7125) the wrong account was associated with the game. [Grumpypuppy](https://heroeslounge.gg/user/view/1028) is the account that got associated, but it should have been [grumpypuppy_1](https://heroeslounge.gg/user/view/8762).

I believe this is happens when the player is a sub player, but there are multiple accounts registered with a similar battlenet name. The below code is responsible for finding the sub player Heroes Lounge account, but `array_search` returns the first match and thus doesn't take into account that there could be multiple eligible matches.

```PHP
$allKey = array_search(strtolower($playerDetails["m_name"]), $this->allSlothNames);
```

I don't know if there's any additional data to be able to 100% identify a player or that this is the best that we can do.

Astraeus-:
Resolves #148 for the most part.

This will try to match substitute players by their `heroesprofile_id`, before falling back to matching by their battletag name. As our `heroesprofile_id` data is quite complete, this should prevent wrong sloths being assoicated with games.

This already wasn't a problem for players part of the team on the website , because we initially limit our search to website team members. (And it was really unlikely that people had the same battletag name on the same team)

### DIFF
--- a/plugins/rikki/heroeslounge/classes/replayparsing/ReplayParsing.php
+++ b/plugins/rikki/heroeslounge/classes/replayparsing/ReplayParsing.php
@@ -319,9 +319,9 @@ class ReplayParsing
                 $participation->hero = $hero;
                 if ($playerDetails["m_teamId"] == 0) {
                     $participation->team_id = $this->match->teams[$firstReplayTeam]->id;
-                    $subSloth = Sloth::where('heroesprofile_id', $playerDetails["m_toon"]["m_id"])->first();
-                    if (isset($subSloth)) {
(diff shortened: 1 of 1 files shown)


Astraeus-:
Made this the primary check for identifying the sloth to associate with the game.

Astraeus-:
I'll create another pr if I do find time to do a refactor. Besides the duplication I noticed some more things that could be updated.

PR #151, opened 2019-09-19
Reduce notifications for users on multiple teams

### LINKED ISSUE #121: Think about notifications
The number of notifications for people in multiple teams is rather high - for example, users are notified when one of their teams is not participating in a season that they are participating in with another team. Filter out unnecessary notifications.

Astraeus-:
Fixes #121 I reduced the number of notifications users would receive when they're part of multiple teams and we're checking their participation.

I was unsure if I should limit the `checkRoster` notification to only show when the team is registered for a season. 
If the above is desired, how about adding an additional statement that we'll fill their team when sign ups close.

### DIFF
--- a/plugins/rikki/heroeslounge/classes/helpers/NotificationHelper.php
+++ b/plugins/rikki/heroeslounge/classes/helpers/NotificationHelper.php
@@ -166,11 +166,11 @@ class NotificationHelper
     {
         $registerableSeasons = Seasons::where('is_active', 1)->where('reg_open', 1)->where('region_id', $sloth->region_id)->get();
         foreach ($registerableSeasons as $season) {
-            if ($season->free_agents->contains($sloth->id)) {
+            if ($sloth->seasons->contains($season)) {
                 continue;
             }
 
-            if ($sloth->isSignedUpForSeason($season)) {
+            if ($sloth->isInTeamParticipatingInSeason($season)) {
                 continue;
             }
 
@@ -185,7 +185,7 @@ class NotificationHelper
             if ($sloth->isCaptain()) {
                 $retVal[] = [
                     'type' => 'warning',
-                    'message' => 'You arent\'t listed to participate in '.$season->title.'. You can register as a free agent or you can sign up the team you\'re captain of on the season page!',
+                    'message' => 'You arent\'t listed to participate in '.$season->title.'. You can register as a free agent or you can sign up your team on the season page!',
                     'entity' => $season->toArray()
                 ];
             } else {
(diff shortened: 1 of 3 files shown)


plugins/rikki/heroeslounge/classes/helpers/NotificationHelper.php
```diff
@@ -163,30 +164,43 @@ private static function generateApplicationMessages($a, $sloth, &$retVal)
...
     public static function checkParticipation($sloth, &$retVal)
     {
-        foreach ($sloth->teams as $key => $team) {
-            $s = Seasons::where('is_active', 1)->where('reg_open', 1)->where('region_id', $team->region_id)->whereDoesntHave('teams', function ($q) use ($sloth, $team) {
-                $q->where('team_id', $team->id);
-            })->get();
-            NotificationHelper::generateParticipationMessages($s, $team, $sloth, $retVal);
+        $registerableSeasons = Seasons::where('is_active', 1)->where('reg_open', 1)->where('region_id', $sloth->region_id)->get();
+        foreach ($registerableSeasons as $season) {
+            if ($season->free_agents->contains($sloth->id)) {
```
Fabian-Sommer: It would be better to just do $sloth->seasons->contains($season) - should be significantly faster

plugins/rikki/heroeslounge/classes/helpers/NotificationHelper.php
```diff
@@ -163,30 +164,43 @@ private static function generateApplicationMessages($a, $sloth, &$retVal)
...
+                continue;
+            }
+
+            if ($sloth->isSignedUpForSeason($season)) {
```
Fabian-Sommer: This function probably wants to be renamed into $isInTeamParticipatingInSeason()

plugins/rikki/heroeslounge/classes/helpers/NotificationHelper.php
```diff
@@ -163,30 +164,43 @@ private static function generateApplicationMessages($a, $sloth, &$retVal)
...
+    private static function generateParticipationMessages($season, $sloth, &$retVal)
     {
-        foreach ($s as $season) {
-            if ($sloth->isCaptainOfTeam($team)) {
+        if ($sloth->teams->count() > 0) {
+            if ($sloth->isCaptain()) {
                 $retVal[] = [
                     'type' => 'warning',
-                    'message' => 'Your team ' . $team->title . ' isn\'t listed to participate in '.$season->title.'. You can sign up your team on the season page!',
+                    'message' => 'You arent\'t listed to participate in '.$season->title.'. You can register as a free agent or you can sign up the team you\'re captain of on the season page!',
```
Fabian-Sommer: Instead of "the team you are captain of" just "your team"

Issue #154, opened 2019-09-29
Reduce number of Heroes Profile ID Sync calls


Astraeus-:
In light of #153 I'd like to reduce our overall number of calls to the Heroes Profile API. I'd like to limit our weekly scheduled fetching of Heroes Profile ID's to accounts without a listed ID or remove the scheduled process entirely as I'm quite certain that Heroes Profile ID's don't change.

For new accounts we retrieve their Heroes Profile ID data on account creation. (When this fails, it would be probably be nice to have the fallback of attempting to sync accounts without a Heroes Profile ID every week). We'd need to add logic for when a user's battletag is changed, to fetch the new matching Heroes Profile ID.

PR #155, opened 2019-10-01
Heroesprofile api update

### LINKED ISSUE #153: Heroes Profile API changes
Heroes Profile is going to be changing how to consume their API soon and we'll need to be ready for it.

Information about the changes of use can be found [here](https://api.heroesprofile.com). I'll take care of getting a new token for us to use, but we basically need to account for their ratelimit.

Example of response header when making a request:
```
 cache-control: no-cache, private 
 content-encoding: gzip 
 content-length: 3226 
 content-type: application/json 
 date: Sun, 29 Sep 2019 19:02:09 GMT 
 server: Google Frontend 
 status: 200 
 vary: Accept-Encoding 
 x-cloud-trace-context: 5a7c5e5c5a4f90ca9586fad7aaef58d6;o=1 
 x-ratelimit-limit: 60 
 x-ratelimit-remaining: 59 
```

As these changes are also new for them, Heroes Profile is open to feedback about the limits and if they'd need to be tweaked.

Astraeus-:
Closes #153 In my testing with a subset of our user-base this handled their rate-limit. I increased the `set_time_limit` to 60 to be on the safe side to not let the script time out, as they reset the permitted requests every minute.

We'll need to update our authorization token, as we were required to generate a new one. I can send this to you when you deploy.

### DIFF
--- a/plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
+++ b/plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
@@ -11,12 +11,8 @@ class IDFetcher
         $sloths = SlothModel::all();
 
         foreach ($sloths as $sloth) {
-            set_time_limit(60);
-            $throttleTime = IDFetcher::fetchIDHeroesProfile($sloth);
-
-            if ($throttleTime > 0) {
-                sleep($throttleTime);
-            }
+            set_time_limit(30);
+            IDFetcher::fetchIDHeroesProfile($sloth);
         }
     }
 
@@ -89,11 +85,14 @@ class IDFetcher
         }
         $sloth->save();
 
-        // Returns the throttle time to wait for the next request.
+        // Check if we've hit the rate-limit and retry the request.
         if (array_key_exists('retry-after', $headers)) {
-            return $headers['retry-after'][0];
-        }
+            $throttleTime = $headers['retry-after'][0];
 
-        return 0;
+            if ($throttleTime > 0) {
+                sleep($throttleTime);
+                IDFetcher::fetchIDHeroesProfile($sloth);
+            }
+        }
     }
 }
--- a/plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
+++ b/plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
@@ -55,7 +55,7 @@ class IDFetcher
         
         $sloth->heroesprofile_id = null;
 
-        $url = 'https://www.heroesprofile.com/API/Profile/?&api_key=' . AuthCode::getHeroesProfileKey() . '&battletag=' . urlencode($battletag) . '&region=' . $region;
+        $url = 'https://api.heroesprofile.com/api/Player?mode=json&api_token=' . AuthCode::getHeroesProfileKey() . '&battletag=' . urlencode($battletag) . '&region=' . $region;
 
         $ch = curl_init($url);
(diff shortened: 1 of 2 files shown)


plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
```diff
@@ -11,8 +11,12 @@ public static function fetchIDs()
         $sloths = SlothModel::all();
 
         foreach ($sloths as $sloth) {
-            set_time_limit(30);
-            IDFetcher::fetchIDHeroesProfile($sloth);
+            set_time_limit(60);
```
Fabian-Sommer: Please assume that our server might just ignore these

plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
```diff
@@ -11,8 +11,12 @@ public static function fetchIDs()
...
+            $throttleTime = IDFetcher::fetchIDHeroesProfile($sloth);
+
+            if ($throttleTime > 0) {
+                sleep($throttleTime);
+            }
         }
     }
```
Fabian-Sommer: This function is probably not actually needed anymore

plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
```diff
@@ -55,22 +59,41 @@ public static function fetchIDHeroesProfile($sloth)
...
         $output = curl_exec($ch);
 
         curl_close($ch);
         if ($output != "null") {
             $data = json_decode($output, true);
             if ($data != null) {
-                if (array_key_exists("blizz_id", $data[0])) {
-                    $sloth->heroesprofile_id = $data[0]["blizz_id"];
+                if (array_key_exists("blizz_id", $data)) {
+                    $sloth->heroesprofile_id = $data["blizz_id"];
```
Fabian-Sommer: I assume that these are changes in their API

plugins/rikki/heroeslounge/classes/mmr/MMRFetcher.php
```diff
@@ -21,8 +25,12 @@ public static function updateMMRs()
...
 
         foreach ($sloths as $sloth) {
-            set_time_limit(30);
-            MMRFetcher::updateMMRHeroesProfile($sloth);
+            set_time_limit(60);
+            $throttleTime = MMRFetcher::updateMMRHeroesProfile($sloth);
+
+            if ($throttleTime > 0) {
+                sleep($throttleTime);
+            }
```
Fabian-Sommer: We will have to see if this works on our server. I had bad experiences with increasing time limit in the past

plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
Astraeus-: That's correct, they changed this in their response

Astraeus-:
I need to re-think the throttle logic a bit for this, as I realized that in the current state it'll drop the request that it failed on instead of retrying it after our limit resets.

plugins/rikki/heroeslounge/classes/hotslogs/IDFetcher.php
Astraeus-: Will update this in a different PR for #154

Astraeus-:
I tested the following with a large number of requests and didn't run into any timeout issues myself even at 30, so I changed it back to that. I added retry logic for the throttled request as well.

PR #158, opened 2019-10-16
Fix discord_id checks

Astraeus-:
Cleaned up the `discord_id` checks to actually work as intended.

When a `discord_tag` gets updated, we should always fetch the new `discord_id` (This was already happening, because `!isset($this->discord_id)` always returns true)

We now also no longer needlessly make failed calls for updating region roles for users that we don't have a `discord_id` for. I debated adding code to get the `discord_id` if that is the case, but since we already try to sync up users every week I left it out.

### DIFF
--- a/plugins/rikki/heroeslounge/components/slothaccount/update.htm
+++ b/plugins/rikki/heroeslounge/components/slothaccount/update.htm
@@ -228,7 +228,7 @@
                     <div class="input-group-btn">
                         <button style="padding:7px; border:solid #868e96 1px; border-left:none;"
                                 data-request="{{ __SELF__ }}::onSyncDiscord"
-                                title="Sync Discord tag" class="close sync-btn" type="button">
+                                title="Fetch new Discord tag from Discord" class="close sync-btn" type="button">
                             <i class="sync-icon fa fa-refresh" style="color: black;" aria-hidden="true"></i>
                         </button>
                     </div>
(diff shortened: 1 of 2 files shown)


Fabian-Sommer:
This looks good. Could you also check under which conditions it is possible for players to change their tag in the first place? I think it is not as intended right now

Astraeus-:
Currently users can enter a new discord tag if we don't have `discord_id` data on them and updates the `discord_id` data. Otherwise they press the sync button and it'll retrieve their updated info.

According to the twig documentation empty strings evaluate to false, so it works as intended however I can make the check more clear if you'd like that.

On the backend moderators can always update a discord tag, which tries to update the `discord_id`.

Fabian-Sommer:
Could you try to make the function (syncing FROM discord) a bit more clear for users? Because right now it looks like one should enter the new discord tag, and then confirm by pressing "Update Discord tag".

Astraeus-:
Currently the only people that can enter a Discord tag are users that we don't have a `discord_id` for. So all they can do is press sync.

Users that we don't have a `discord_id` for can update their Discord tag and save, which will try to fetch a `discord_id`. At that point the above will also become the case for them.

Fabian-Sommer:
I am aware of the process. Please design a intuitive interface for that - Moving the sync button to the left and changing the text might suffice.

Astraeus-:
Keeping it at the right looked the nicest in my opinion. 

Changed the sync button to only appear when we have a `discord_id` for the user, to remove confusion to what the user needs to do. Also updated the wording to `sync` from `update` to be more clear as to what is going to happen.

Fabian-Sommer:
Sync has the same problem as update as in that it is not intuitive that we fetch the data directly from discord, rather than using the input right next to the button.
Please change the tooltip to "Fetch new Discord tag from Discord".

Issue #140, opened 2019-08-17
Users can not update season participation


Astraeus-:
The below snippet is supposed to trigger when users try to update their free agent status from the front-end. However we no longer use `sloth->team_id` since sloths can be a part of multiple teams. As a result users can not remove themselves from the FA list.

https://github.com/Fabian-Sommer/HeroesLounge/blob/917027845b9e36b2a6a3dd00760290076aa5f3db/plugins/rikki/heroeslounge/components/SlothAccount.php#L520-L522

PR #160, opened 2019-10-20
Handle free agent assignment and removal

### LINKED ISSUE #126: Remove free agents when they join a team
When a user that is signed up as free agent for a season joins a team that is signed up for that season, or when a team that the user is in signs up for that season, the user should be removed from that seasons free agent list

Astraeus-:
Fixes #126 and #140.

Users will be removed from the Free Agent list if they join a team signed up to a season with open registration or one of their teams signs up for a season.

Also allows users signed up as a Free Agent, but not participating with a team to remove their sign up from a season with open registration.

### DIFF
--- a/plugins/rikki/heroeslounge/components/SlothAccount.php
+++ b/plugins/rikki/heroeslounge/components/SlothAccount.php
@@ -279,26 +279,27 @@ class SlothAccount extends UserAccount
 
     public function onParticipationSave()
     {
-        try {
-            foreach ($this->seasons as $season) {
-                if ($season->current_round == 0) {
-                    $val = post('part_seas_'.$season->id);
-                    if (isset($val) && !$season->free_agents->contains($this->sloth->id)) {
-                        $season->free_agents()->attach($this->sloth->id);
-                        Discord\RoleManagement::UpdateUserRole("PUT", $this->sloth->discord_id, "FreeAgent");
-                        Flash::success('You will participate in '.$season->title.'!');
-                    } elseif (isset($val) == false) {
-                        $season->free_agents()->detach($this->sloth->id);
-                        Discord\RoleManagement::UpdateUserRole("DELETE", $this->sloth->discord_id, "FreeAgent");
-                        Flash::error('You will not paricipate in '.$season->title.' - Sad to see you go!');
-                    }
+        foreach ($this->seasons as $season) {
+            if ($season->reg_open == 1) {
+                if ($this->sloth->isInTeamParticipatingInSeason($season)) {
+                    Flash::error('You can not update your '.$season->title.' participation, because you are already participating with a team!');
+                    return Redirect::refresh(); 
+                }
+
+                $participation = post('part_seas_'.$season->id);
+                if (isset($participation) && !$season->free_agents->contains($this->sloth->id)) {
+                    $season->free_agents()->attach($this->sloth->id);
+                    Discord\RoleManagement::UpdateUserRole("PUT", $this->sloth->discord_id, "FreeAgent");
+                    Flash::success('You will participate in '.$season->title.'!');
+                } elseif (isset($participation) == false) {
+                    $season->free_agents()->detach($this->sloth->id);
+                    Discord\RoleManagement::UpdateUserRole("DELETE", $this->sloth->discord_id, "FreeAgent");
+                    Flash::error('You will not participate in '.$season->title.' - Sad to see you go!');
                 }
             }
-        } catch (Exception $e) {
-            Flash::error($e->getMessage());
-        } finally {
-            return Redirect::refresh();
         }
+
+        return Redirect::refresh();
     }
 
     public function onUpdateGeneral()
@@ -518,9 +519,8 @@ class SlothAccount extends UserAccount
         if ($sloth->region_id == 2 && isset($data['server_preference'])) {
             $sloth->server_preference = $data['server_preference'];
         }
-        if ($sloth->team_id == 0) {
-            $this->onParticipationSave();
-        }
+
+        $this->onParticipationSave();
         $sloth->save();
 
         Flash::success('Role updated successfully!');
--- a/plugins/rikki/heroeslounge/components/SlothAccount.php
+++ b/plugins/rikki/heroeslounge/components/SlothAccount.php
@@ -277,31 +277,6 @@ class SlothAccount extends UserAccount
         }
     }
 
-    public function onParticipationSave()
-    {
-        foreach ($this->seasons as $season) {
-            if ($season->reg_open == 1) {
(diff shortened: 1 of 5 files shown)


plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -279,26 +279,27 @@ public function onRegister()
...
-                        $season->free_agents()->attach($this->sloth->id);
-                        Discord\RoleManagement::UpdateUserRole("PUT", $this->sloth->discord_id, "FreeAgent");
-                        Flash::success('You will participate in '.$season->title.'!');
-                    } elseif (isset($val) == false) {
-                        $season->free_agents()->detach($this->sloth->id);
-                        Discord\RoleManagement::UpdateUserRole("DELETE", $this->sloth->discord_id, "FreeAgent");
-                        Flash::error('You will not paricipate in '.$season->title.' - Sad to see you go!');
-                    }
+        foreach ($this->seasons as $season) {
+            if ($season->reg_open == 1) {
```
Fabian-Sommer: You can leave away the == 1

plugins/rikki/heroeslounge/components/ViewApps.php
```diff
@@ -118,6 +120,15 @@ public function onAccept()
                     //a sloth should not participate in two different teams in the same season or tournament
                     if ($this->sloth->mayJoinTeam($team)) {
                         $this->sloth->teams()->add($team);
+
+                        $seasons = Season::all();
+                        foreach ($seasons as $season) {
+                            if ($season->reg_open == 1 && $season->teams->contains($team) && $season->free_agents->contains($this->sloth)) {
```
Fabian-Sommer: == 1

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -279,26 +279,27 @@ public function onRegister()
...
+                    Flash::error('You will not participate in '.$season->title.' - Sad to see you go!');
                 }
             }
-        } catch (Exception $e) {
-            Flash::error($e->getMessage());
-        } finally {
-            return Redirect::refresh();
         }
+
+        return Redirect::refresh();
```
Fabian-Sommer: General comment: We can just get rid of the entire "Sign up on your profile" concept

Astraeus-: So would we still want users to be able to remove their participation from a season on their profile? Or would we want this to also be removed and added somewhere else?

Fabian-Sommer: Ideally, all of the signing up / remove participation should be on the season / playoff page

Astraeus-: Updated the free agent participation changes to be on season pages. I did keep a listing of participation on the profile page for legacy reasons.

I didn't change anything for playoffs, as we don't allow free agent sign ups there anyway.

plugins/rikki/heroeslounge/components/ViewApps.php
```diff
@@ -118,6 +120,15 @@ public function onAccept()
                     //a sloth should not participate in two different teams in the same season or tournament
                     if ($this->sloth->mayJoinTeam($team)) {
                         $this->sloth->teams()->add($team);
+
+                        $seasons = Season::all();
+                        foreach ($seasons as $season) {
+                            if ($season->reg_open && $season->teams->contains($team) && $season->free_agents->contains($this->sloth)) {
+                                $season->free_agents()->remove($this->sloth);
```
Fabian-Sommer: Why did you change detach to remove?

plugins/rikki/heroeslounge/components/slothaccount/update.htm
```diff
@@ -286,29 +285,35 @@ <h3>Server Preference</h3>
...
+            <div>
+                <label>
+                    <input type="checkbox" name="part_seas_{{season.id}}" {% if checked %} checked {% endif %} disabled>
+                    {{season.title}}
+                </label>
+            </div>
+            {% set checked = '0' %}
+        {% endfor %}
 
         <!-- END PARTICIPATION -->
```
Fabian-Sommer: Please remove this from this page

plugins/rikki/loungeviews/components/ParticipationOverview.php
```diff
@@ -99,9 +105,25 @@ public function onSlothSignup()
...
         }
     }
+
+    public function onSlothRemoveSignUp()
+    {
+        $this->user = Auth::getUser();
+        $this->season = Seasons::find($_POST['season_id']);
+        if ($this->user != null) {
+            $sloth = $this->user->sloth;
+            if ($sloth != null && $sloth->region_id == $this->season->region_id && $this->season->free_agents->contains($sloth)) {
```
Fabian-Sommer: All of these checks should be unnecessary.

plugins/rikki/loungeviews/components/participationoverview/default.htm
```diff
@@ -30,10 +30,20 @@ <h1>{{__SELF__.season.title}} Participation</h1>
                 class="btn btn-sm btn-secondary center"
                 data-request-data="season_id: {{__SELF__.season.id}}"
                 data-request="{{ __SELF__ }}::onSlothSignup">
-            Sign up as a free agent!
+            Sign up as a free agent
         </button>
         <br><br>
     </div>
+    {% elseif __SELF__.user and __SELF__.signedUp and __SELF__.user.sloth.region_id == __SELF__.season.region_id %}
```
Fabian-Sommer: Why do you check for the right region id here?

plugins/rikki/loungeviews/components/participationoverview/default.htm
```diff
@@ -30,10 +30,20 @@ <h1>{{__SELF__.season.title}} Participation</h1>
...
+    <div>
+        No longer want to particpate as a free agent?<br>
```
Fabian-Sommer: Is there something like 'You are signed up as a free agent' on this page? If not, please add it before this.

plugins/rikki/heroeslounge/components/ViewApps.php
Astraeus-: Since we use the adding equivalent when a sloth signs up for a season. I can use detach if you prefer though.

Fabian-Sommer: Remove is fine if it works

Astraeus-: I tested it locally and it worked as expected

Astraeus-:
Noticed in final testing that checking signup was general and not specific to free agent, so made sure that the user is actually a free agent to allow them to sign up or no longer participate

PR #161, opened 2019-11-05
Add map losses and map score to division table

### LINKED ISSUE #120: Show map losses on division standings
Division standings should be sorted by
1. Match wins
2. Map wins
3. Map losses
To reflect the tiebreaker rules that are relevant for end-of-season standings.
Standings should also show map losses

Bonus:
4. Average KD (total kills / total kills)

Astraeus-:
Fixes #120 I think? I did not account for head to head results in the sorting, but this should be an improvement nonetheless. I also reordered the division table to look like tables in other sports and added map scores. I can remove the map scores if they're too much clutter. 

I tried playing around with`match_count` as sort weight, but I didn't find one that worked, so I removed it. 

Example of new table from [Season 8 division 4](https://heroeslounge.gg/eu-season-8/division-4):

image

### DIFF
--- a/plugins/rikki/loungeviews/components/DivisionTable.php
+++ b/plugins/rikki/loungeviews/components/DivisionTable.php
@@ -42,17 +42,24 @@ class DivisionTable extends ComponentBase
                 //calculate game wins
                 foreach ($this->teams as $team) {
                     $team->game_wins = 0;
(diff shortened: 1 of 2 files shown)


Astraeus-:
Removed the map losses column from the division standings.
New look:

image

PR #162, opened 2019-11-05
Rework registration and updating sloths

Astraeus-:
I noticed that there's currently no way to get a new `heroesprofile_id` when a user changes their battletag or a way for moderators to force a resync if a user does happen to get out of sync.

This fix fetches Heroes Profile data for newly registering users or users that get their battletag changed (Also triggers if the sloth's `heroesprofile_id` is null and a different field is updated).

I tested all of this locally and it worked as expected for new registrations and user updates.

### DIFF
--- a/plugins/rikki/heroeslounge/components/SlothAccount.php
+++ b/plugins/rikki/heroeslounge/components/SlothAccount.php
@@ -221,11 +221,6 @@ class SlothAccount extends UserAccount
                 Auth::login($user);
             }
 
-
-            /*
-            * Attach a Sloth to the user
-            */
-
             $sloth = new SlothModel;
             $sloth->user = $user;
             $sloth->title = $user->username;
--- a/plugins/rikki/heroeslounge/components/SlothAccount.php
+++ b/plugins/rikki/heroeslounge/components/SlothAccount.php
@@ -26,8 +26,6 @@ use Rikki\Heroeslounge\Models\SlothRole;
 use Rikki\Heroeslounge\Models\Timeline;
 use Rikki\Heroeslounge\classes\Mailchimp\MailChimpAPI;
 use Rikki\Heroeslounge\Models\Region as Region;
-use Rikki\Heroeslounge\classes\hotslogs\IDFetcher;
-use Rikki\Heroeslounge\classes\mmr\MMRFetcher;
 
 class SlothAccount extends UserAccount
 {
@@ -236,9 +234,6 @@ class SlothAccount extends UserAccount
 
             $sloth->save();
             $this->user = $user;
-
-            IDFetcher::fetchIDHeroesProfile($sloth);
-            MMRFetcher::updateMMRHeroesProfile($sloth);
 
             /*
               Assign EU or NA role on Discord based on region_id.
--- a/plugins/rikki/heroeslounge/components/SlothAccount.php
+++ b/plugins/rikki/heroeslounge/components/SlothAccount.php
@@ -223,35 +223,25 @@ class SlothAccount extends UserAccount
 
 
             /*
-            * Make sure the user gets a Sloth attached
+            * Attach a Sloth to the user
             */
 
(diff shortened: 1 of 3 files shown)


Astraeus-:
Will take a deeper look at this before it's completely read and address the life-cycle methods and resolve the issues mentioned in devops for updating battletag and discord information in the backend.

Astraeus-:
I cleaned up the sloth model code a lot more than I anticipated. I moved most of the post-registration or sloth update code on to the sloth model. I also addressed the life-cycle bug when trying to update a user's Discord tag.

I was unsure if we'd only want MMR to be immediately fetched on registration or also after battletag updates. I've kept it only after registration for now, to stay consistent with what we currently have (Since we do an MMR sync once per week).

Should be ready for deployment.

plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -225,38 +223,25 @@ public function onRegister()
 
 
             /*
-            * Make sure the user gets a Sloth attached
+            * Attach a Sloth to the user
```
Fabian-Sommer: Delete this useless comment

plugins/rikki/heroeslounge/models/Sloth.php
```diff
@@ -127,15 +127,56 @@ public static function getFromUser($user)
...
+        }
 
-        $user->sloth = $sloth;
+        MMRFetcher::updateMMRHeroesProfile($this);
 
-        return $sloth;
+        $this->_saveTimelineEntry('Sloth.Created');
+        if (!empty($this->team_id)) {
+            $this->_saveTimelineEntry('Sloth.Joins.Team');
+        }
```
Fabian-Sommer: This is outdated and can be deleted

plugins/rikki/heroeslounge/models/Sloth.php
```diff
@@ -127,15 +127,56 @@ public static function getFromUser($user)
...
+            region_id = 2: NA
+        */
+        if ($this->region_id == 1) {
+            Discord\RoleManagement::UpdateUserRole("PUT", $this->discord_id, "EU");
+        } else if ($this->region_id == 2) {
+            Discord\RoleManagement::UpdateUserRole("PUT", $this->discord_id, "NA");
+        }
 
-        $user->sloth = $sloth;
+        MMRFetcher::updateMMRHeroesProfile($this);
```
Fabian-Sommer: Have you tested that this works?

plugins/rikki/heroeslounge/models/Sloth.php
```diff
@@ -127,15 +127,56 @@ public static function getFromUser($user)
...
+    {
+        if ($this->isDirty('battle_tag') || $this->heroesprofile_id == null) {
+            IDFetcher::fetchIDHeroesProfile($this);
+        }
+    }
+
+    public function beforeDelete()
+    {
+        $this->_saveTimelineEntry('Sloth.Deleted');
     }
```
Fabian-Sommer: Delete this, this can never ever trigger

plugins/rikki/heroeslounge/models/Sloth.php
```diff
@@ -127,15 +127,56 @@ public static function getFromUser($user)
...
+                Discord\RoleManagement::UpdateUserRole("DELETE", $this->discord_id, "EU");
+                Discord\RoleManagement::UpdateUserRole("PUT", $this->discord_id, "NA");
+            }
+        }
+    }
+
+    public function beforeSave()
+    {
+        if ($this->isDirty('battle_tag') || $this->heroesprofile_id == null) {
+            IDFetcher::fetchIDHeroesProfile($this);
```
Fabian-Sommer: Have you tested that this works with the new value?

Astraeus-:
I addressed the feedback. I tested all of the flows locally and they worked as expected.

PR #163, opened 2019-11-23
Fix and enhance match reschedule notification logic

Astraeus-:
This moves the match reschedule logic to the model event so that it can trigger consistently on match reschedules. E.g: In the case that someone was to reschedule a match from the back-end.

Also fixes a bug that a notifications would still appear when all casters are denied.
When we're missing `discord_id` data from a caster, we'll reference their website username instead of an empty mention.

### DIFF
--- a/plugins/rikki/heroeslounge/models/Match.php
+++ b/plugins/rikki/heroeslounge/models/Match.php
@@ -166,14 +166,6 @@ class Match extends Model
         });
     }
 
-    public function getAcceptedOrPendingCasters()
-    {
-        return $this->casters
-        ->filter(function ($item) {
-            return $item->pivot->approved == 0 || $item->pivot->approved == 1;
-        });
-    }
-
     public function getAppliedCasterIds()
     {
         return $this->casters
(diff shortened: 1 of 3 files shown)


plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -162,6 +166,14 @@ public function getAcceptedCasters()
...
     }
 
+    public function getAcceptedOrPendingCasters()
+    {
+        return $this->casters
+        ->filter(function ($item) {
+            return $item->pivot->approved == 0 || $item->pivot->approved == 1;
+        });
+    }
+
```
Fabian-Sommer: You can split this into two functions (one of which exists already), or just do this in the function itself (although the first approach would be better)

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -205,6 +217,37 @@ public function beforeUpdate()
...
+                $newDate = new DateTime($this->wbp, new DateTimeZone(TimezoneHelper::defaultTimezone()));
+                $oldDate = new DateTime($this->original['wbp'], new DateTimeZone(TimezoneHelper::defaultTimezone()));
+
+                if ($this->teams[0]->region_id == 1) {
+                    $newDate->setTimezone(new DateTimeZone('Europe/Berlin'));
+                    $oldDate->setTimezone(new DateTimeZone('Europe/Berlin'));
+                } else if ($this->teams[0]->region_id == 2) {
+                    $newDate->setTimezone(new DateTimeZone('America/Los_Angeles'));
+                    $oldDate->setTimezone(new DateTimeZone('America/Los_Angeles'));
+                }
```
Fabian-Sommer: nit: just determine the timezone in the if, then apply it outside

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -205,6 +217,37 @@ public function beforeUpdate()
...
+
+                // Create our information message to inform assigned / pending casters.
+                $notificationString = "The match between " . $this->teams[0]->title . " and " . $this->teams[1]->title . " has been rescheduled from " . $oldDate->format('d M Y H:i T') . " to " . $newDate->format('d M Y H:i T') . "\n";
```
Fabian-Sommer: Please check that there are two teams before you access the arrays

Astraeus-:
Example of the new look:
image

PR #165, opened 2020-02-18
Improves performance for team hero and map stats

Astraeus-:
Optimized the database query to fetch all the data in one go, instead of fetching individual pieces during the later loops. 

Observed about a 100x performance increase locally for api endpoints for both. The front-end didn't have this issue. 

I didn't observer a noticeable increase for sloth statistics, so I didn't change those.


plugins/rikki/loungestatistics/classes/statistics/Statistics.php
```diff
@@ -230,7 +231,7 @@ public static function calculateMapStatisticsForSloth($gameParticipations, $seas
     }
 
     public static function calculateMapStatisticsForTeam($team, $season)
-    {
+    {
```
Fabian-Sommer: nit: delete whitespace

PR #166, opened 2020-02-19
Optimizations

Astraeus-:
More query optimizations, the most notable being on DivisionTable.

### DIFF
--- a/plugins/rikki/loungeviews/components/DivisionTable.php
+++ b/plugins/rikki/loungeviews/components/DivisionTable.php
@@ -23,6 +23,7 @@ class DivisionTable extends ComponentBase
     public $startIndex = 0;
     public $length =  null;
     public $showScore = false;
+    public $matches = null;
 
     public function onRender()
     {
@@ -34,6 +35,7 @@ class DivisionTable extends ComponentBase
                                     whereNull('rikki_heroeslounge_teams.deleted_at')->
                                     orderBy('win_count', "DESC")->orderBy('free_win_count', "ASC")->orderBy('bye', "ASC")->orderBy('match_count', "DESC")->get();
 
+            $this->matches = $div->matches()->with('teams')->get();
             if ($this->property('showScore')) {
                 //used for playoffs
                 //we want teams that since then disbanded as well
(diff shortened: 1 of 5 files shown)


plugins/rikki/loungeviews/components/TimelineEntries.php
```diff
@@ -107,7 +107,8 @@ public function onRender()
                     break;
                 case 'division':
                     $timeline = new Collection([]);
-                    Timeline::with('divisions', 'matches.teams.divisions', 'teams.divisions', 'sloths.teams', 'teams.logo')->orderBy('created_at', 'desc')->chunk(100, function ($someTimelines) use ($id, $timeline) {
+                    Timeline::with('divisions', 'matches.teams.divisions', 'teams.divisions', 'sloths.teams.divisions', 'teams.logo')->orderBy('created_at', 'desc')
+                    ->chunk(100, function ($someTimelines) use ($id, $timeline) {
```
Fabian-Sommer: When you are splitting a line, the new lines should start 2*4=8 spaces further inwards than the original line. (Also, the first line is still super-long, 80 characters is generally a good guideline)

plugins/rikki/loungeviews/components/DivisionTable.php
```diff
@@ -44,7 +46,7 @@ public function onRender()
                     $team->game_wins = 0;
                     $team->game_losses = 0;
                 }
```
Fabian-Sommer: You should initialize the variable closer to where it is used the first time

plugins/rikki/loungeviews/components/TimelineEntries.php
```diff
@@ -107,7 +107,10 @@ public function onRender()
                     break;
                 case 'division':
                     $timeline = new Collection([]);
-                    Timeline::with('divisions', 'matches.teams.divisions', 'teams.divisions', 'sloths.teams', 'teams.logo')->orderBy('created_at', 'desc')->chunk(100, function ($someTimelines) use ($id, $timeline) {
+                    Timeline::with('divisions', 'matches.teams.divisions', 'teams.divisions', 'sloths.teams.divisions', 'teams.logo')
+                            ->orderBy('created_at', 'desc')->chunk(100, function ($someTimelines) use ($id, $timeline) {
+
+
```
Fabian-Sommer: nit: Never two empty lines at once (and here I think none since we are just starting a block of code)

PR #168, opened 2020-04-25
Minor fixes and clean up

Astraeus-:
Fixes match games nav getting rendered (specifically on mobile) when there are no games yet for a match.

Fixes timeline entries to be have their images and text hidden at the start before bouncing in.

Updates the reschedule match logic to prevent matches being scheduled before the it's creation date (As this will likely never happen)

### DIFF
--- a/plugins/rikki/heroeslounge/components/UpdateMatch.php
+++ b/plugins/rikki/heroeslounge/components/UpdateMatch.php
@@ -88,17 +88,19 @@ class UpdateMatch extends ComponentBase
                 $wbp = new DateTime($date, new DateTimeZone($timezone));
                 $wbp->setTimezone(new DateTimeZone(TimezoneHelper::defaultTimezone()));
                 $match->wbp = $wbp->format('Y-m-d H:i:s');
-                if ($match->tbp != null && Carbon::parse($match->wbp) < Carbon::parse($match->tbp) && Carbon::parse($match->wbp) > Carbon::parse($match->created_at)) {
+                if ($match->tbp != null && Carbon::parse($match->wbp) < Carbon::parse($match->tbp) && Carbon::parse($match->wbp) >= Carbon::now()) {
                     $match->save();
-                    Flash::success('Match has been successfully rescheduled for '.$date);
+                    $wbp->setTimezone(new DateTimeZone($timezone));
+                    Flash::success('Match has been successfully rescheduled for '. $wbp->format('d M Y H:i') );
                 } else {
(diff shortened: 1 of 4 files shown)


Astraeus-:
Implemented your feedback and formatted the new date in the success response.

plugins/rikki/heroeslounge/components/UpdateMatch.php
```diff
@@ -88,17 +88,19 @@ public function onRescheduleMatch()
...
-                    $matchCreate->setTimezone(new DateTimeZone($timezone));
-
-                    Flash::error('The match has to be played between ' . $matchCreate->format('d M Y H:i') . ' and ' . $tbp->format('d M Y H:i'));
+                    if (Carbon::parse($match->wbp) > Carbon::parse($match->tbp)) {
+                        $tbp = new DateTime($match->tbp, new DateTimeZone(TimezoneHelper::defaultTimezone()));
+                        $tbp->setTimezone(new DateTimeZone($timezone));
+                        Flash::error('The match has to be played before ' . $tbp->format('d M Y H:i'));
+                    } else {
+                        $wbp->setTimezone(new DateTimeZone($timezone));
+                        Flash::error('The new schedule time ' . $wbp->format('d M Y H:i') . ' can not be in the past');
```
Fabian-Sommer, line 102: nit: cannot

Fabian-Sommer: This can be implemented without nesting this much. You can do
if (...) {
 ...
 return
}
if (...)
...

plugins/rikki/heroeslounge/components/UpdateMatch.php
```diff
@@ -97,10 +97,11 @@ public function onRescheduleMatch()
                         $tbp = new DateTime($match->tbp, new DateTimeZone(TimezoneHelper::defaultTimezone()));
                         $tbp->setTimezone(new DateTimeZone($timezone));
                         Flash::error('The match has to be played before ' . $tbp->format('d M Y H:i'));
-                    } else {
-                        $wbp->setTimezone(new DateTimeZone($timezone));
-                        Flash::error('The new schedule time ' . $wbp->format('d M Y H:i') . ' can not be in the past');
+                        return;
```
Fabian-Sommer, line 100: This is still the same nesting as before, and it also returns the wrong thing

Astraeus-: I think I misunderstood originally, will fix.

Astraeus-:
I think I rearranged the validation logic properly in a more legible format using short circuiting. Verified everything locally to make sure it's all working.

PR #173, opened 2020-05-22
Dependencies update

Astraeus-:
Updated dependencies:
- Bootstrap: v4.0.0-beta => v4.5.0
- Chartjs:	v2.7.0 => Removed
- jquery: v1.12.4 => v3.5.1
- jquery-migrate: v1.4.1 => v3.3.0
- jquery-ui: v1.12.1 => v1.12.1
- jquery smartmenus: v1.0.0 => v.1.1.0
- parallax: v1.4.2 => v.1.5.0
- jquery datetimepicker: v1.3.3 => v2.5.20
- datatables: v1.10.16 => v1.10.21
- popper: v1.12.4 => v1.16.1 (LTS)

Fixed functionality that broke due to dependency updates.


Fabian-Sommer:
Please elaborate on "functionality"

Astraeus-:
I guess "functionality" isn't the exact right term here, but I basically had to add a new css rule for the navigation header dropdown, since they would otherwise not work anymore. The datatable styling also changed a little, so I had to change the rules there.

Everything else worked as expected in my local testing.

Fabian-Sommer:
We should not be rolling updates unless there is a very compelling reason to. We don't have any testing in place to ensure that nothing breaks.

PR #174, opened 2020-05-25
Discordtag sync fix

Astraeus-:
No longer relies on a user being present in the Heroes Lounge Discord to be able to sync their Discord tag. Also added `CURLOPT_FAILONERROR` to error requests with status codes >=400.


plugins/rikki/heroeslounge/classes/discord/Attendance.php
```diff
@@ -118,19 +118,16 @@ public static function CheckIndividualAttendance($discordId)
...
+      curl_setopt($ch, CURLOPT_FAILONERROR, true);
       $output = curl_exec($ch);
 
       if (curl_errno($ch)) {
           return false;
       }
 
-      $memberData = json_decode($output, true);
-
       curl_close($ch);
```
Fabian-Sommer, line 128: nit: No newlines in 123, 127, but keep the one after curl_close. Ditto below.

PR #175, opened 2020-05-25
Fix mailchimp during registration

Astraeus-:
Now passes in an up to date user to the Mailchimp api instead of the incomplete reference we had leftover from registration.

During registration we create and save a new sloth and attach the user we get from our auth provider. However after saving our sloth the `$sloth->user` user reference is unaware of what sloth it's been attached to. 

Fetching from the database get us an up to date usermodel instance, which is updated.

### DIFF
--- a/plugins/rikki/heroeslounge/components/SlothAccount.php
+++ b/plugins/rikki/heroeslounge/components/SlothAccount.php
@@ -18,6 +18,7 @@ use RainLab\User\Models\Settings as UserSettings;
 use Exception;
 
 use RainLab\User\Components\Account as UserAccount;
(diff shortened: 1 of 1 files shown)


plugins/rikki/heroeslounge/components/SlothAccount.php
```diff
@@ -230,13 +231,12 @@ public function onRegister()
             $sloth->region_id = $data['region_id'];
             $sloth->save();
 
-            $this->user = $sloth->user;
-
+            $user = User::where('username', $sloth->title)->first();
```
Fabian-Sommer, line 234: nit: firstOrFail

PR #176, opened 2020-05-25
Add scheduled job for HeroUpdater

Astraeus-:
I went for Tuesdays, since game patches tend to happen on Wednesday, so it'd likely give our image source the most time to update their information, since I don't think it's automated on their end.

I saw that we have a scheduled task for fetching heroprotocol, do we still need this considering we try to update the protocol when parsing fails?


Fabian-Sommer:
We want to fetch heroprotocol since when parsing fails = a user is waiting.

PR #177, opened 2020-05-26
Navigation rework

Astraeus-:
Started on a refactor of the header navigation. The following key points were mentioned by Zalz to keep in mind in order of priority:

- (1) It should work on mobile.
- (2) It should be completely editable from the backend, preferably without knowing html
- (3) It should work without a mouse (somehow, I cannot tab into any of the current drop downs)
 - (4) Actually removing layers of dropdowns. 2 people are only a very small fraction of the userbase.

Point (1) I'll be testing once I'm happy with the final changes, but so far everything still works as expected.

To address point (2) I moved the Method Mayhem and the Events dropdown menus into partials. This would allow editing on the backend on the `CMS` tab, but would require documentation, which I can write. I did try to use the built-in menus/pages, but our dropdown menus are too custom to be able to make good use of these.

Point (3) hasn't been looked at yet.

For point (4) I've only done the amateur series archive. I also looked into the events archive, but to clean up the division-s links would require adding relations between qualifiers and their season and there's some inconsistent naming for ARAM leagues, so I've held off for now.

Preview of the new archive dropdown:
image

Preview of the new season overview page:
image

I'd love to hear some feedback if you think this is the right direction, or answer any questions you may have.


Astraeus-:
I personally am not entirely happy with this archive sub-menu change, as I don't think it actually solves the problem that the underlying menu is growing too large. I'd rather create a dedicated archive page, listing all the seasons, with links to their divisions and playoffs.

Fabian-Sommer:
Go for the archive page then

Astraeus-:
The archive submenu has been replaced with a link to the new archive page.

I implemented a similar design to our FAQ page with expanding cards. All season cards are originally closed on page render and only one card can be open at a time.
image

Astraeus-:
I've also implemented an archive page for events.
image

This page contains no automation, but instead relies on wrapping a static page that can be edited on the backend pages menu. Below are required steps after deployment that are optional and required for this to work.


Optional changes:
- Streamline naming for ARAM league seasons
- Streamline naming for Division S divisions/playoffs
- Navigation partials should probably be .gitignored, since they will be edited to change our production navigation?

Required changes:
- Create a page named "events-archive"
- Paste in my local page implementation
- Associate division S qualifiers with their seasons

Astraeus-:
To summarize the changes the PR adds/allows the following:
- Adds overview page for the current season types that we have (Amateur, Division S and Aram).
- Moves the amateur season archive to a dedicated page based on it's original logic.
- Moves the events archive to a dedicated page that can be edited from the `pages` backend.
- Updated the navigation archive submenus to be links to their archive page.
- Allows editing the (current) Method Mayhem and Events dropdowns from the `CMS -> partials` backend.

When it's ready to be deployed, I'll make sure the required changes are ready in production.

It worked on mobile in my local testing and since I've only made use of similar elements to what we had, I'd be surprised if it will cause problems in production. I looked into navigating the menus without a mouse, but I couldn't get it to work. I suspect it's due to our use of submenus implemented with some custom code as there's no official bootstrap support for it on our version.

themes/HeroesLounge-Theme/pages/season/archive.htm
```diff
@@ -0,0 +1,30 @@
+title = "Amateur Season Archive"
```
Fabian-Sommer, line 1: Amateur Series Archive

themes/HeroesLounge-Theme/pages/season/archive.htm
```diff
@@ -0,0 +1,30 @@
...
+url = "/season/archive"
+layout = "plain"
+is_hidden = 0
+==
+
+use Rikki\Heroeslounge\Models\Season;
+
+function onStart()
+{
+  $this['previous_seasons'] = Season::where('type', 1)->with('divisions', 'playoffs')->where('is_active', 0)->orderBy('created_at','desc')->get();
```
Fabian-Sommer, line 11: order by region first, creation date second

themes/HeroesLounge-Theme/pages/season/archive.htm
```diff
@@ -0,0 +1,30 @@
...
+}
+==
+<h1>Amateur Season Archive</h1>
```
Fabian-Sommer, line 14: see above

themes/HeroesLounge-Theme/partials/navigation/events.htm
```diff
@@ -0,0 +1,16 @@
+<li class="nav-item dropdown">
+  <a class="nav-link dropdown-toggle" id="divsnav" data-toggle="dropdown" aria-haspopup="true"
+    aria-expanded="false">Events</a>
+  <ul role="menu" class="dropdown-menu pull-right">
+    <li class="nav-item dropdown-submenu">
+      <a class="dropdown-toggle dropdown-item" data-toggle="dropdown">[EU] Offmeta Maps</a>
```
Fabian-Sommer, line 6: This event is over, so move it to archive

themes/HeroesLounge-Theme/partials/navigation/events.htm
```diff
@@ -0,0 +1,16 @@
+<li class="nav-item dropdown">
```
Fabian-Sommer, line 1: backend/cms is not there to be used to edit anything. So please don't use partials unless it is to avoid code duplication. All this achieves is tech debt where it is completely unclear where the navigation is living.

Astraeus-:
Addressed feedback and created the events archive page in production in preparation for deployment without any of the optional changes or division qualifier associations.

plugins/rikki/loungeviews/components/navigation/default.htm
```diff
@@ -91,9 +91,9 @@
                         </ul>
                     </li>
 
-                    {% partial '/navigation/featured' %}
+                    {% partial '@featured' %}
 
-                    {% partial '/navigation/events' %}
+                    {% partial '@events' %}
```
Fabian-Sommer, line 96: This is still a pointless partial

themes/HeroesLounge-Theme/partials/season/overview.htm
```diff
@@ -0,0 +1,28 @@
...
+  {% endfor %}
+</ul>
+{% endif %}
+
+{% if season.playoffs|length > 0 %}
+  {% if season.type == 3 %}
+  <h2 class="mt-1"> Stages </h2>
+  {% elseif season.type == 2 %}
+  <h2 class="mt-1"> Tournaments </h2>
+  {% else %}
```
Fabian-Sommer, line 17: nit: season.type == 1

Fabian-Sommer: Also, order them 1, 2, 3

plugins/rikki/loungeviews/components/navigation/default.htm
```diff
@@ -63,8 +63,8 @@
                             <hr>
```
Fabian-Sommer, line 63: nit: At line 49: We should sort divisions there by title, like in the archive

PR #180, opened 2020-06-04
Talent image fetch fix

Astraeus-:
There were two problems with my previous implementation.

1. The logic for updating partial talents that we created during replay parsing should have come first, since instead of updating this partial talent we were creating a new one. This change will try to update partial talents before creating new talents. This means that for some talents we're going to have duplicate entries, but that should not be a problem. (These entries need to stay, because they may have associations in `rikki_heroeslounge_gameparticipation_talent`).

2. The logic assumed that talents for which we have all the information have an image stored on disk, however in practice this does not always seem to be the case. The new logic will always call `setTalentImage`, but checks if we already have an entry on disk.

### DIFF
--- a/plugins/rikki/heroeslounge/classes/heroes/HeroUpdater.php
+++ b/plugins/rikki/heroeslounge/classes/heroes/HeroUpdater.php
@@ -116,7 +116,6 @@ class HeroUpdater
             mkdir($talent_image_path, 0777, true);
         }
 
-        // Check if the file already exists and if it's larger than 500 bytes.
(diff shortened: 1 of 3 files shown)


plugins/rikki/heroeslounge/classes/heroes/HeroUpdater.php
```diff
@@ -111,6 +116,11 @@ public static function setTalentImage($talent_model, $icon_url)
             mkdir($talent_image_path, 0777, true);
         }
 
+        // Check if the file already exists and if it's larger than 500 bytes.
```
Fabian-Sommer, line 119: nit: This comment does not add anything that isnt obvious in the line below, so just delete it.

PR #182, opened 2020-06-09
Filter upcoming matches to list playoff match times per round

Astraeus-:
Filters playoff matches to only list a single entry for each round. This assumes that there is only a single playoff on a day.

New look:
image

### DIFF
--- a/plugins/rikki/heroeslounge/components/UpcomingMatches.php
+++ b/plugins/rikki/heroeslounge/components/UpcomingMatches.php
@@ -15,6 +15,7 @@ use Request;
 use Carbon\Carbon;
 use Log;
 use RainLab\User\facades\Auth;
+use October\Rain\Support\Collection;
 
 class UpcomingMatches extends ComponentBase
 {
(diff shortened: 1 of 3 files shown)


plugins/rikki/heroeslounge/components/UpcomingMatches.php
```diff
@@ -84,25 +85,39 @@ public function collectMatches($timezoneName, $id)
...
                 } elseif ($this->idApp == 'pending') {
-                    $matches = $myEntity->castMatches()->where('rikki_heroeslounge_match_caster.approved', '=', '0')->where('winner_id', null)->orderBy('wbp', 'asc')->where('wbp', '>=', Carbon::today())->where('wbp', '<=', Carbon::today()->addDays($daysInFuture))->get();
+                    $matches = $myEntity->castMatches()->where('rikki_heroeslounge_match_caster.approved', '=', '0')->where('winner_id', null)->orderBy('wbp', 'asc')->orderBy('id', 'asc')->where('wbp', '>=', Carbon::today())->where('wbp', '<=', Carbon::today()->addDays($daysInFuture))->get();
                 } else {
-                    $matches = $myEntity->castMatches()->orderBy('wbp', 'asc')->where('winner_id', null)->where('wbp', '>=', Carbon::today())->where('wbp', '<=', Carbon::today()->addDays($daysInFuture))->get();
+                    $matches = $myEntity->castMatches()->where('winner_id', null)->orderBy('wbp', 'asc')->orderBy('id', 'asc')->where('wbp', '>=', Carbon::today())->where('wbp', '<=', Carbon::today()->addDays($daysInFuture))->get();
```
Astraeus-, line 97: This is mostly reordering the query builder statements to be consistent, but I also added a secondary `orderBy` on id, since otherwise the two playoff rounds with the same start time could be out of order.E.g: Round 2 would be listed before Round 1 in the calendar.

plugins/rikki/heroeslounge/components/UpcomingMatches.php
```diff
@@ -84,25 +85,39 @@ public function collectMatches($timezoneName, $id)
...
                     return Carbon::parse($match->wbp)->setTimezone($timezoneName)->format('d-M-y');
                 }
-            );
+            )->map(function($dateMatches) {
+                $listedRounds = new Collection([]);
+
+                return $dateMatches->filter(function($match) use($listedRounds) {
+                    if (!$match->playoff) {
+                        return true;
+                    } else {
```
Astraeus-, line 111: I've currently done the filtering for all playoff types, but perhaps it's better to only filter out the playoffs that don't belong to a season, so everything stays as is for end of season playoffs?

PR #184, opened 2020-06-13
Timeline entries

Astraeus-:
Adds new timeline entries and create entries for roster changes through the backend.

New entries:
- Deactivation due to failing to schedule on time
- Deactivation due to failing to upload results on time
- Deactivation by moderator on the backend
- Activation by moderator on the backend

Adds calls for existing entries:
- `Sloth.Joins.Team` created when sloth is added to team on backend
- `Sloth.Left.Team` created when sloth is removed from team on backend


Fabian-Sommer:
I don't like the timeline as it is right now for a bunch of reasons, for sloths the match history does pretty much convey all the info in a better way, for teams the info is very incomplete. We can merge the fixes to sloth joining/leaving team, the added entries should not be exposed to the frontend as part of the timeline.

PR #192, opened 2021-07-17
Adjust tbp Date for last round

Kaktushose:
Fixed the bug that got reported in [#feedback_and_suggestions](https://discord.com/channels/200267155479068672/494650946438168577/865913175067852830)

image

### DIFF
--- a/plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
+++ b/plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
@@ -21,6 +21,12 @@ class Swiss
         $timeoffset = ($pairings[0][0]->region_id - 1)*36000;
         $schedule_date = date('Y-m-d H:i:s', strtotime('next sunday 21:55')+$timeoffset);
         $tbp = date('Y-m-d H:i:s', strtotime('+1 weeks sunday 21:55')+$timeoffset);
+        
+        // if it's the last round, games must be played until next friday (+5 days)
+        if ($div->season->current_round == 10) {
+	        $tbp = date('Y-m-d H:i:s', strtotime('+5 days sunday 21:55')+$timeoffset);
+        } 
+        
         $created_at = date('Y-m-d H:i:s' , strtotime('now'));
         foreach ($pairings as $pairing) {
             $this->createMatch($pairing, $div, $div->season->current_round, $created_at, $schedule_date, $tbp);
--- a/plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
+++ b/plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
@@ -22,9 +22,9 @@ class Swiss
         $schedule_date = date('Y-m-d H:i:s', strtotime('next sunday 21:55')+$timeoffset);
         $tbp = date('Y-m-d H:i:s', strtotime('+1 weeks sunday 21:55')+$timeoffset);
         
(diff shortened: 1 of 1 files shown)


plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
```diff
@@ -21,6 +21,12 @@ public function createMatches($pairings, $div) {
         $timeoffset = ($pairings[0][0]->region_id - 1)*36000;
         $schedule_date = date('Y-m-d H:i:s', strtotime('next sunday 21:55')+$timeoffset);
         $tbp = date('Y-m-d H:i:s', strtotime('+1 weeks sunday 21:55')+$timeoffset);
+        
+        // if it's the last round, games must be played until next friday (+5 days)
+        if ($div->season->current_round == 10) {
```
Astraeus-, line 26: Any reason you're not using the `$currentRound` property used throughout this class?

Additionally, this would blow up when a season does not have 10 rounds.

Kaktushose: the createMatch method used it neither, so I assumed it's not initialized at that point. 

Actually good point, this would screw everything up. But: Has there ever be a season with more than 10 rounds?

Fabian-Sommer: There are seasons with != 10 rounds, e.g. Nexus Rumble.
You can compare to season->round_length which has the number of rounds in the season

plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
```diff
@@ -21,6 +21,12 @@ public function createMatches($pairings, $div) {
         $timeoffset = ($pairings[0][0]->region_id - 1)*36000;
         $schedule_date = date('Y-m-d H:i:s', strtotime('next sunday 21:55')+$timeoffset);
         $tbp = date('Y-m-d H:i:s', strtotime('+1 weeks sunday 21:55')+$timeoffset);
+        
+        // if it's the last round, games must be played until next friday (+5 days)
```
Fabian-Sommer, line 25: Nit: "If it's the last round, games must be played until Friday instead of Sunday."

plugins/rikki/heroeslounge/classes/matchmaking/Swiss.php
```diff
@@ -21,6 +21,12 @@ public function createMatches($pairings, $div) {
...
+        if ($div->season->current_round == 10) {
+	        $tbp = date('Y-m-d H:i:s', strtotime('+5 days sunday 21:55')+$timeoffset);
```
Fabian-Sommer, line 27: '+1 weeks friday' is correct here - I think +5 days sunday would be equivalent to next sunday (for code that runs on Monday), i.e. ~7 days instead of ~12 days

Kaktushose: I'm not familiar with php, but I just tested it in PHP sandbox. Both `+5 days sunday` and `+1 weeks friday` have the same result (in this case Friday, 30.07). However, I think that `+1 weeks friday` is easier to understand and more consistent

Fabian-Sommer (approved):
LGTM
Let me know if you want me to merge & deploy this

Kaktushose:
sure

PR #198, opened 2023-02-21
Fix: No image for Hanamura Temple #196

### LINKED ISSUE #196: No image for Hanamura Temple
An image for the Hanamura Temple battleground is currently not being loaded, likely due to the fact that we do not have an image to load for this battleground.

Example match where the battleground is played https://heroeslounge.gg/match/view/19110

image

CSNWEB:
Fix #196. Didn't tested this locally since I do not have the data, but if there is no typo in the name of the file this should work fine.


Fabian-Sommer (approved):
LGTM

PR #199, opened 2023-02-28
Set matchmaking dates in local timezone (Fix #199)

CSNWEB:
Makes sure the deadlines are always at 23:55 in the timezone of the region independent of DST changes.

I did test the code part I changed independently and it seems to work fine, I didn't try to run the match scheduling, not sure if that would have worked locally.


CSNWEB:
Fixes #199

Fabian-Sommer (approved):
LGTM

Issue #203, opened 2023-12-28
Nexus Rumble MMR tool


TransparentCharlie:
A tool page for looking up a list of teams signed up for an event.
- Team Name
- Team MMR
- Captains Heroes Lounge name
- Captains Discord ID

This would remove the need for anyone assisting with the Nexus rumble needing database access.

PR #204, opened 2023-12-29
Add additional team info to playoffs

CSNWEB:
Implements #203 as discussed in Discord by adding additional columns to the playoff teams list


Fabian-Sommer (approved):
LGTM - Instead of chunk(), slice() would be more appropriate, but this is in line with the implementation above, which is more important. Thanks!
I can implement the actual tool page requested in #203 with this.

PR #205, opened 2023-12-29
Basic Single Playoff permissions

CSNWEB:
This is a rather basic implementation of a limited playoff / matches admin. They can edit playoffs they were assigned to as admin, but not the divisions (groups). The ability to create new playoffs is hidden in the UI, but theoretically they can circumvent that right now and create one (and lose access to it immediately). Same for matches, they can edit matches of the playoff they are assigned to. The playoff and match list are limited to their assigned playoffs. I would argue this will probably be sufficient, since most people we trust enough to get access at all won't try too hard to circumvent it, and if they do they get banned. Its not like they can actually really fuck anything up, but YMMV.

### DIFF
--- a/plugins/rikki/heroeslounge/models/Match.php
+++ b/plugins/rikki/heroeslounge/models/Match.php
@@ -15,6 +15,8 @@ use Carbon\Carbon;
 use DateTime;
 use DateTimeZone;
 use Log;
+use BackendAuth;
+
 /**
  * Model
  */
@@ -86,6 +88,27 @@ class Match extends Model
             'table' => 'rikki_heroeslounge_match_channel'
         ]
     ];
+
+    protected static function boot()
+    {
+        parent::boot();
+        // Check if BackendAuth exists, as its missing in some environments like console commands / cronjobs
+        if (class_exists('BackendAuth')) {
+            $user = BackendAuth::getUser();
+            // Check if the user really exists
+            if (is_object($user)) {
+                // Add scope that only shows playoffs that the user has access to
+                static::addGlobalScope('limit_playoff_access', function ($builder) use ($user) {
+                    $builder->whereHas(
+                        'playoff.backend_users',
+                        function ($backendUserBuilder) use ($user) {
+                            $backendUserBuilder->where('backend_users.id', $user->id);
+                        }
+                    );
+                });
+            }
+        }
+    }
 
     public function listParticipatingTeamsForBackend($fieldname, $value, $formData)
     {
--- a/plugins/rikki/heroeslounge/models/Playoff.php
+++ b/plugins/rikki/heroeslounge/models/Playoff.php
@@ -6,6 +6,7 @@ use Rikki\Heroeslounge\Models\Match;
 use Rikki\Heroeslounge\Classes\Helpers\TimezoneHelper;
 use Carbon\Carbon;
 use Log;
+use BackendAuth;
 
 class Playoff extends Model
 {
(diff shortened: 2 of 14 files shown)


plugins/rikki/heroeslounge/plugin.yaml
```diff
@@ -122,6 +126,9 @@ permissions:
     rikki.heroeslounge.casterscheduling:
         tab: 'rikki.heroeslounge::lang.plugin.name'
         label: 'Appove casters, assign channels'
+    rikki.heroeslounge.single_playoff:
+        tab: 'rikki.heroeslounge::lang.plugin.name'
+        label: 'Manage Single playoffs and its matches'
```
Fabian-Sommer, line 131: nit: 'single playoff'
Also, I would probably use the word 'individual' over 'single', since one admin could be assigned to multiple playoffs.

Fabian-Sommer (changes requested):
Thanks, this looks pretty good. Just a few comments

plugins/rikki/heroeslounge/plugin.yaml
CSNWEB: So I would go with individual_playoffs? And Manage individual playoffs @Fabian-Sommer?

plugins/rikki/heroeslounge/models/Playoff.php
```diff
@@ -36,8 +37,36 @@ class Playoff extends Model
...
+            'otherKey' => 'backend_user_id',
+            'table' => 'rikki_heroeslounge_backend_users_playoff'
+        ]
     ];
 
+    protected static function boot()
+    {
+        parent::boot();
+        // Check if BackendAuth exists, as its missing in some environments like console commands / cronjobs
+        if (class_exists('BackendAuth')) {
```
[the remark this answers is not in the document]
CSNWEB: But I did fix the comment and scope name

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -87,6 +89,30 @@ class Match extends Model
...
 
+    protected static function boot()
+    {
+        parent::boot();
+        // Check if BackendAuth exists, as its missing in some environments like console commands / cronjobs
+        if (!class_exists('BackendAuth')) {
+            return
+        }
+        $user = BackendAuth::getUser();
+            // Check if the user really exists
```
Fabian-Sommer, line 100: Nit: Wrong indent

plugins/rikki/heroeslounge/models/Playoff.php
```diff
@@ -36,8 +37,39 @@ class Playoff extends Model
...
 
+    protected static function boot()
+    {
+        parent::boot();
+        // Check if BackendAuth exists, as its missing in some environments like console commands / cronjobs
+        if (!class_exists('BackendAuth')) {
+            return
+        }
+        $user = BackendAuth::getUser();
+            // Check if the user really exists
```
Fabian-Sommer, line 57: Nit: Wrong indent

plugins/rikki/heroeslounge/models/Match.php
```diff
@@ -87,6 +89,30 @@ class Match extends Model
...
+        if (!is_object($user)) {
+            return
+        }
+
+        // Add scope that only shows matches that the user has access to
+        static::addGlobalScope('limit_match_access', function ($builder) use ($user) {
+            $builder->whereHas(
+                'playoff.backend_users',
+                function ($backendUserBuilder) use ($user) {
+                    $backendUserBuilder->where('backend_users.id', $user->id);
```
Fabian-Sommer, line 110: Maybe I am reading this wrong, but the way I understand this is that for all matches that have a playoff assigned, and that playoff has backend_users, the playoff is hidden if the user is not allowlisted there. The problem is that existing users that should have access don't see it. Have you tested that users that already have access don't lose it (and, for completeness, that the frontend isn't impacted, but I think this should work out).
If I am correct here, a solution would be to only apply this filter if the user has the individual_playoffs permission.

plugins/rikki/heroeslounge/plugin.yaml
```diff
@@ -25,6 +26,7 @@ navigation:
                 icon: icon-tag
                 permissions:
                     - rikki.heroeslounge.season
+                    - rikki.heroeslounge.individual_playoffs
```
Fabian-Sommer, line 29: For my own understanding: Access is granted for users with any of these, right?

plugins/rikki/heroeslounge/models/Match.php
CSNWEB: Hm I did have two browsers open, and thought I checked that the first time I pushed. (I obviously did not test the second time sorry, otherwise the semicolons would have been there). But the code clearly does what you think it does and should only apply this if they don’t have general access. The frontend will not be impacted since the backend auth and user will be undefined.

plugins/rikki/heroeslounge/plugin.yaml
CSNWEB: Yes

CSNWEB:
Finally fixed the issues and tested it again, seems to work now. Thanks @Fabian-Sommer

